PR #3, opened 2022-11-15
User model POC

fynsta:
Adds a proof of concept for the internal representation of a user (for type safety and methods that we can define on the model).

We will need to write similar methods for every table we use (I have no idea how to reduce that duplication)


istzen (approved):

MarkFalke (approved):

Eddie-42 (approved):

Showcas (commented):
sieht gut aus

PR #12, opened 2022-11-24
Login process

fynsta:
Finally done! Please try it out locally and verify that everything works as expected :) 
(Note that you need to open the ResetPassword-Mail on the emulated device to try out the deeplink).

Please ask questions if you don't understand some code!

Includes:
- Welcome Page
- Login Page
- Register Page
- Forgot Password Page
- Reset Password Page

Also renames Settings -> Account (more general) and adds a Logout-Button on that page (without any styling for now, that will be part of another User Story)

Authored by @Eddie-42 and @fynsta


Eddie-42 (approved):

Showcas (approved):

lib/welcome/pages/register_page.dart
```diff
@@ -0,0 +1,205 @@
...
+    });
+    await Future.delayed(const Duration(seconds: 2));
+    setState(() {
+      _state = ButtonState.idle;
+    });
+  }
+
+  void onSuccess() async {
+    ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
+      content: Text("Account created. Please log in"),
```
istzen, line 106: can we write here something like "please confirm your mail, then you can login", so that it's clear that you have to confirm the mail before a login is possible.

lib/welcome/pages/register_page.dart
```diff
@@ -0,0 +1,205 @@
...
+      late final AuthResponse res;
+      try {
+        setState(() {
+          _state = ButtonState.loading;
+        });
+        res = await supabaseClient.auth.signUp(
+            password: passwordController.text, email: emailController.text);
+      } on AuthException {
+        fail();
+        showSnackBar("Something went wrong.");
```
istzen, line 62: Do we want to tell the user if the mail already exists?

android/app/src/main/AndroidManifest.xml
```diff
@@ -24,6 +25,16 @@
...
+
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <!-- Accepts URIs that begin with YOUR_SCHEME://YOUR_HOST -->
+                <data
+                android:scheme="io.supabase.flutter"
+                android:host="reset-callback" />
+            </intent-filter>
```
istzen, line 37: i think the deep-link for login-callback is missing

lib/welcome/pages/login_page.dart
```diff
@@ -0,0 +1,166 @@
...
+        ScaffoldMessenger.of(context).showSnackBar(SnackBar(
+          content: Text(text),
+        ));
+      }
+    } else {
+      fail();
+    }
+  }
+
+  void fail() async {
```
istzen, line 89: can we refactor this method? because it's the same everywhere it's used.

lib/welcome/pages/forgot_password_page.dart
```diff
@@ -0,0 +1,120 @@
...
+      supabaseClient.auth.resetPasswordForEmail(emailController.text,
+          redirectTo: kIsWeb ? null : 'io.supabase.flutter://reset-callback/');
+
+      await Future.delayed(const Duration(seconds: 1));
+      setState(() {
+        _state = ButtonState.success;
+      });
+      await Future.delayed(const Duration(seconds: 2));
+
+      onMailSent();
```
istzen, line 71:
```suggestion
      if (mounted) {
        Navigator.of(context).pop();
      }
```

you can also check if widget is 'mounted', then you can use context in an async method.

lib/welcome/pages/register_page.dart
fynsta: @Eddie-42 do we want to use snackbar for this at all? Maybe we should just redirect to some static page that tells you to look into your mails? Because you will get logged in via the deeplink anyways

Eddie-42: I don't know how other apps do this, but having a separate page seems like a hassle. It seems a little awkward for me for a user to be redirected to a page that they've never seen without any user interaction, I can imagine people getting confused how this happened and where in the app structure they are right now and, more importantly, I would be a bit annoyed by getting sent to an entirely non-functional page. I would suggest having the snackbar just read "Please log in via the e-mail link" or something and not redirecting anywhere or just to automatically open the default e-mail app?

istzen: I think the snackbar is enough for now. We can see when we do our User study if that is unclear, then we can change it later. 
Maybe we can redirect to the welcome page?

lib/welcome/pages/login_page.dart
fynsta: sadly, we don't really think there's a good way to do that :( Guess we'll have to accept the duplication.

lib/welcome/pages/register_page.dart
fynsta: we'll have to ask the AG about that, not sure... Added a TODO comment for now

istzen (approved):

PR #18, opened 2022-12-03
Line length 120

fynsta:
Changed the line length for dart to 120 (seems like a good value to me, what do you think?) Then I ran `dart format . --line-length=120`.
We definitely have a lot more space to work with now!

I changed the value in the VSCode configuration so it should be applied automatically there, but you should also change it in Android Studio if you use it (search for "Dart line length" in the settings).


Eddie-42 (approved):

istzen (approved):

Showcas (approved):

fynsta:
ping @MarkFalke

MarkFalke (approved):

PR #21, opened 2022-12-07
Drive detail

Eddie-42:
Add a detailed page for drives. It contains
- a small overview at the top,
- a detailed timeline which shows who enters and who leaves at every waypoint,
- a list of riders,
-  and a "Delete" button.

Note: There is a constructor DriveDetailPage.fromDrive, which enables us to load some details in advance and have them shown on the DriveDetailPage without pesky loading time.

Additionally, we made it possible for the ride and drive models to contain its riders and rides as proper objects and not just IDs. This can be used with supabase calls joining the respective tables as shown in the code. There is also a method in the drive model for determining the max number of seats in a drive.

### DIFF
--- a/lib/drives/pages/drives_page.dart
+++ b/lib/drives/pages/drives_page.dart
@@ -1,5 +1,8 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_app/drives/models/drive.dart';
 import 'package:flutter_app/drives/pages/create_drive_page.dart';
+import 'package:flutter_app/drives/pages/drive_detail_page.dart';
+import 'package:flutter_app/util/supabase.dart';
 
 class DrivesPage extends StatefulWidget {
   const DrivesPage({super.key});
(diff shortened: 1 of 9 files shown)


Eddie-42:
For the person who is responsible for unifying the timelines designs in the future, it might be relevant that the code for building a FixedTimeline without a builder has been removed in the latest commit to make way for the new overview. You can just look in the commit before that for a good example of how to use that

fynsta (approved):

lib/drives/pages/drives_page.dart
```diff
@@ -19,14 +22,20 @@ class _DrivesPageState extends State<DrivesPage> {
...
-          );
-        },
+        onPressed: onPressed,
         backgroundColor: Theme.of(context).colorScheme.primary,
         child: const Icon(Icons.add),
       ),
     );
   }
+
+  void onPressed() async {
```
istzen, line 32: did you want that to direct to the detailview and change it before the merge? because right now it's not possible to check whether the deletebutton works.

lib/rides/models/ride.dart
```diff
@@ -14,6 +15,7 @@ class Ride extends Model {
 
   final int driveId;
   final int riderId;
+  final Profile? rider;
```
istzen, line 18: if we have the rider for a ride, do we really need the riderId additionally?
and we could also add the driver but thats maybe in another PR

fynsta: The problem ist that the `rider` is only present if you do the join in the SQL query. But `rider_id` is always present, and needed for the Json export.

Yes of course adding `driver` will be a logical step when we need it.

MarkFalke (approved):

PR #23, opened 2022-12-12
add flutter version manager

istzen:
with the flutter version manager (fvm) it's easy to change the flutter versions and make sure that we are all working with the same one. 
I think at the moment it's not important, but since i had a different version installed than all of you what lead to the problems with my last PR i installed it so that this does not happen again. (for more information see [here](https://fvm.app/)).

@fynsta thanks for finding fvm.


fynsta:
just fyi the link in the description is wrong

Showcas (approved):

PR #30, opened 2022-12-21
Remove search Detail

fynsta:
Please look at this, but I will just merge it to unblock Philipp.

Removes SearchDetail and makes RideDetail support preview rides.


Eddie-42 (approved):

istzen (commented):
Why did you remove the id from previewFromDrive?

fynsta:
because we already have the status to signify that it's a preview. Using the id as well would be doubled.

Showcas (approved):

PR #32, opened 2022-12-23
Address suggestions 2

fynsta:
This fixes most of the open TODOs from the last PR:

- Also shows train stations (might be good meeting places)
- Saves search history in local storage (those storage results are ranked by how often they were used and when the last usage was). The storage is done by user id, so that you won't get the results of someone else.
- Removes duplicates that are returned by MOTIS
Also moves MotisHandler to more general AddressSuggestionHandler.

The ranking algorithm needs some work, do you have any ideas for a good ranking there?


lib/util/search/address_search_delegate.dart
```diff
@@ -70,6 +71,10 @@ class AddressSearchDelegate extends SearchDelegate<AddressSuggestion?> {
                 return const Divider();
               },
             );
+          } else {
+            return const Center(
+              child: Text('Please enter more information to get results'),
```
Eddie-42, line 76: Testing at your parent's home with a slow internet connection sometimes works wonders! :) If I type in more than 4 characters quickly, I get this message for a solid amount of time and then fairly briefly get the CircularProgressIndicator. If this happens I think some people might ask themselves how many characters they are supposed to type when in reality they just have a really bad connection. Maybe check if more than 4 characters have been typed in here?

Eddie-42 (changes requested):
When you search for a place you have searched for before, it shows up both as a history and regular suggestion. I think this duplicate should be avoided and history should take precedent

fynsta:
Yes that duplication should definitely be prevented. The only annoying thing is the fact that we might have to deduplicate the motis results (to get 5 different ones), and then the whole array again.

lib/util/search/address_search_delegate.dart
fynsta: Ah yes I can reproduce that. I changed it so the loading indicator is only displayed when it is actually loading.

However, this hints at a few slightly more involved problems, which I'm not sure I want to tackle in this PR:
- The getHistorySuggestions method does not need to be async because I keep the whole suggestions queue in memory. Thus those could already get shown above the other results while it is loading.
But to support that, you would have to restructure that FutureBuilder and it probably leads to code that looks a whole lot worse than how it reads today (not to mention the deduplication of suggestions would be a lot harder).
- We could theoretically detect when the user is not connected to the internet (there is actually a [native package](https://pub.dev/packages/connectivity_plus)) and just not do the network requests when there is no connection. That would at least show the history results in that case. Otherwise it just loads forever.
Of course, there is no good way to detect whether the connected network is actually connected to the internet, but in that case we will just have to do with infinite loading.

Wdyt?

Eddie-42: Honestly, I think the current state could easily be merged and those are cool ideas, but I would push them reeeally far down our prioritized list of Todos. If your connection is bad, actually loading the search results will take a while anyway, and I think that most people know the status bar enough to deduct the reason for infinite loading, so it's not a communication problem either.

Eddie-42 (approved):

lib/util/search/address_search_delegate.dart
fynsta: Definitely agree. I only say this because I absolutely hate apps that start loading and you already know that it will fail in advance because there is just no internet. And I just want to enjoy the offline features, but it does not let me because it is so convinced that it can load stuff. (ahem Spotify ahem)

But yeah not that needed right now.

istzen (approved):

fynsta:
Fyi I also added the possibility to remove suggestions now. Should be helpful if you don't want to see it anymore. That means I will probably remove the counter and just use recency for the ranking.

MarkFalke (approved):

PR #34, opened 2022-12-23
fixes for ride and drive

istzen:
change cancel in drive so that only pending & approved rides get updated 
fix ridesOfUserAtTime
fix previewFromDrive
add overlapsWithTime method to trip

### DIFF
--- a/lib/drives/models/drive.dart
+++ b/lib/drives/models/drive.dart
@@ -65,21 +65,17 @@ class Drive extends Trip {
   List<Ride>? get approvedRides => rides?.where((ride) => ride.status == RideStatus.approved).toList();
   List<Ride>? get pendingRides => rides?.where((ride) => ride.status == RideStatus.pending).toList();
 
-  @override
-  String toString() {
-    return 'Drive{id: $id, from: $start at $startTime, to: $end at $endTime, by: $driverId}';
-  }
-
   static Future<List<Drive>> getDrivesOfUser(int userId) async {
     return Drive.fromJsonList(await supabaseClient.from('drives').select().eq('driver_id', userId));
   }
 
   static Future<Drive?> driveOfUserAtTime(DateTime start, DateTime end, int userId) async {
-    //get all drives of user
-    final List<Drive> drives = await Drive.getDrivesOfUser(userId);
+    //get all upcoming drives of user
+    List<Drive> drives = await Drive.getDrivesOfUser(userId);
+    drives = drives.where((drive) => !drive.cancelled && !drive.isFinished).toList();
     //check if drive overlaps with start and end
     for (Drive drive in drives) {
-      if (drive.startTime.isBefore(end) && drive.endTime.isAfter(start)) {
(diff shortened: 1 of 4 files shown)


lib/drives/models/drive.dart
```diff
@@ -136,6 +132,16 @@ class Drive extends Trip {
   Future<void> cancel() async {
     cancelled = true;
     await supabaseClient.from('drives').update({'cancelled': true}).eq('id', id);
-    await supabaseClient.from('rides').update({'status': RideStatus.cancelledByDriver.index}).eq('drive_id', id);
+    //cancel all pending and approved rides for this drive.
+    await supabaseClient
+        .from('rides')
+        .update({'status': RideStatus.cancelledByDriver.index})
+        .eq('drive_id', id)
+        .or('status.eq.${RideStatus.pending.index},status.eq.${RideStatus.approved.index}');
```
fynsta, line 140: why "or"? I don't get ths

fynsta, line 140: why "or"? I don't get ths

fynsta, line 140: why "or"? I don't get ths

lib/util/trip/trip.dart
```diff
@@ -22,6 +22,10 @@ abstract class Trip extends Model {
   bool get isOngoing => startTime.isBefore(DateTime.now()) && endTime.isAfter(DateTime.now());
 
   bool overlapsWith(Trip other) {
-    return startTime.isBefore(other.endTime) && endTime.isAfter(other.startTime);
+    return overlapsWithTime(other.startTime, other.endTime);
+  }
+
+  bool overlapsWithTime(DateTime start, DateTime end) {
```
fynsta, line 28: this could just be a method overload of "overlapsWith". And it's probably nicer to just give it a time range instead? https://api.flutter.dev/flutter/material/DateTimeRange-class.html

fynsta, line 28: this could just be a method overload of "overlapsWith". And it's probably nicer to just give it a time range instead? https://api.flutter.dev/flutter/material/DateTimeRange-class.html

lib/drives/models/drive.dart
istzen: So that only rides get updated that either are approved or pending.

lib/util/trip/trip.dart
istzen: yes thats a good idea thanks

istzen: i think dart don't allows overloading.

fynsta: ah okay good to know :D

lib/drives/models/drive.dart
fynsta: Ahhhh - So the "or" is just applied between the two conditions inside, and it is connected to the other condition via "and"? That's confusing...

fynsta (approved):

Eddie-42 (approved):

PR #35, opened 2022-12-28
A11y

Eddie-42:
Finally, accessibility!

There are basically 3 ways to improve the user experience of people with disabilities:
* **Large fonts**,
* **Screen readers**, and
* **Sufficient contrast**.

I tested our app with the largest selectable font in Android and it worked fine. In terms of contrast, I think we are already on the right track, but in the future, I think it will be helpful to include automated tests for meeting accessibility guidelines that are, as far as I know, relatively easy to implement. This and more about accessibility in Flutter you can find here: https://docs.flutter.dev/development/accessibility-and-localization/accessibility.

So, this mostly includes changes about the UX with **Screen Readers**. To test this, as the Android Studio emulator doesn't offer Screen Reader support out of the box and most versions (e.g. the Pixel 6 one) don't include the Play Store, you have to download the Android Accessibility Suite APK ([here](https://apkpure.com/android-accessibility-suite/com.google.android.marvin.talkback/download?from=details), for example) and drag-n-drop it into the emulator window.

tl;dr on Screen Readers: You can swipe left and right to go back and forth between semantic elements, or you can tap once on them. You can douple-tap to activate buttons or text fields.

Mostly these improvements are done with ```Semantics```, which bundles its child as one semantic entity. In the simplest cases, ```MergeSemantics``` is sufficient, which tells Flutter that its child has to be read as one entity so that users don't have to swipe right too often. ```Semantics``` is much more powerful. It can do the same thing if ```excludeSemantics``` is set to true, but it can also tell users if entities are buttons and what they are going to do if clicked (this is standard behavior that should always be implemented! Most Flutter buttons do this automatically, but Inkwells are, in my experience, not treated as buttons by Screen Readers and it would be nice for some buttons that are not self-explanatory to assign a ```tooltip```). You can also change the order in which elements are selected when swiping right, which I used to read the Trip FABs before the list of trips, which could be too long.

Note that you will probably need to test if accessibility works for your features every time.

### DIFF
--- a/lib/account/pages/account_page.dart
+++ b/lib/account/pages/account_page.dart
@@ -42,6 +42,12 @@ class _AccountPageState extends State<AccountPage> {
                   }),
             ),
           ),
+          actions: <Widget>[
+            TextButton(
+              child: const Text("Ok"),
+              onPressed: () => Navigator.of(context).pop(),
+            ),
+          ],
         ),
       ),
     );
@@ -68,6 +74,12 @@ class _AccountPageState extends State<AccountPage> {
               ),
             ),
           ),
+          actions: <Widget>[
+            TextButton(
+              child: const Text("Ok"),
+              onPressed: () => Navigator.of(context).pop(),
+            ),
+          ],
         ),
       ),
     );
--- a/lib/account/pages/help_page.dart
+++ b/lib/account/pages/help_page.dart
@@ -84,9 +84,12 @@ class _FAQCardState extends State<FAQCard> {
                         const Divider(
                           thickness: 1,
                         ),
-                        Text(
-                          widget.answer,
-                          textAlign: TextAlign.start,
+                        Semantics(
+                          liveRegion: true,
+                          child: Text(
+                            widget.answer,
+                            textAlign: TextAlign.start,
+                          ),
                         ),
                       ],
                     )
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -218,6 +218,7 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
               position: BadgePosition.topEnd(top: -12),
               child: const Icon(Icons.chat),
             ),
+            tooltip: 'Chat',
           ),
         ],
       ),
@@ -236,18 +237,20 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
   List<Widget> buildCard(Waypoint stop) {
     List<Widget> list = [];
     list.add(
-      Padding(
-        padding: const EdgeInsets.only(bottom: 4),
-        child: Row(
-          mainAxisSize: MainAxisSize.min,
-          children: [
-            Text(
-              localeManager.formatTime(stop.time),
-              style: DefaultTextStyle.of(context).style.copyWith(fontWeight: FontWeight.w700),
-            ),
-            const SizedBox(width: 4.0),
-            Text(stop.place),
-          ],
+      MergeSemantics(
+        child: Padding(
+          padding: const EdgeInsets.only(bottom: 4),
+          child: Row(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              Text(
+                localeManager.formatTime(stop.time),
+                style: DefaultTextStyle.of(context).style.copyWith(fontWeight: FontWeight.w700),
+              ),
+              const SizedBox(width: 4.0),
+              Text(stop.place),
+            ],
+          ),
         ),
       ),
     );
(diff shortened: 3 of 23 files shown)


lib/account/pages/help_page.dart
```diff
@@ -84,9 +84,12 @@ class _FAQCardState extends State<FAQCard> {
                         const Divider(
                           thickness: 1,
                         ),
-                        Text(
-                          widget.answer,
-                          textAlign: TextAlign.start,
+                        Semantics(
+                          liveRegion: true,
```
fynsta, line 88: what does the liveRegion do here? When will this text ever change?

lib/util/loading_button.dart
```diff
@@ -52,7 +52,11 @@ class _LoadingButtonState extends State<LoadingButton> {
       onPressed: widget.onPressed,
       state: widget.state,
       textStyle: TextStyle(
-        color: Theme.of(context).colorScheme.onPrimary,
+        color: widget.state == ButtonState.fail
```
fynsta, line 55: switch statement?

lib/account/pages/account_page.dart
```diff
@@ -68,6 +74,12 @@ class _AccountPageState extends State<AccountPage> {
...
           ),
+          actions: <Widget>[
+            TextButton(
+              child: const Text("Ok"),
+              onPressed: () => Navigator.of(context).pop(),
+            ),
+          ],
         ),
       ),
     );
```
fynsta, line 85: I can't comment on collapsed lines, but the "Sign out" Button does not seem to be clickable currently

lib/util/trip/trip_overview.dart
```diff
@@ -15,28 +15,32 @@ class TripOverview extends StatelessWidget {
...
+                Text(trip.end),
+                Text(
+                  localeManager.formatTime(trip.startTime),
+                  style: DefaultTextStyle.of(context).style.copyWith(fontWeight: FontWeight.w700),
+                )
+              ],
+            ),
           ),
         ),
       ],
```
fynsta, line 46: the seatindicator should probably be changed to have a label "10 of 6 seats", currently it says "10 6 seats" and it's not easy to understand

lib/drives/pages/drive_detail_page.dart
```diff
@@ -238,47 +241,54 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
-                            ? List.generate(action.seats, (index) => icon)
-                            : [
-                                icon,
-                                const SizedBox(width: 2),
-                                Text("x${action.seats}"),
-                              ],
+      Widget container = Semantics(
+        button: true,
+        label:
+            "${action.isStart ? "Pick up" : "Drop-off"} ${action.seats} people associated with ${action.profile.username}.",
```
fynsta, line 247: I like this. But it should probably also announce the start/end of the whole drive? Like it currently only says the place name when you start somewhere

lib/account/pages/help_page.dart
Eddie-42: The text doesn't change, but this makes the Screen Reader read the text automatically when the card is expanded

lib/util/loading_button.dart
Eddie-42: Would that be better? You can only execute statements in switch statements, so I'd have to move that into a method or assign the color before the huge return statement... IMO this is clear enough, but I'll leave that up to you

lib/util/trip/trip_overview.dart
Eddie-42: Yeah right that is a bit weird, I'll change that

lib/drives/pages/drive_detail_page.dart
Eddie-42: I don't really get this, I quite like how it announces the time and place and then who is getting picked up and dropped off, like if you just swipe through the elements I think you get a pretty good idea of the drive

lib/util/loading_button.dart
fynsta: Yeah okay I agree. But maybe move the functionality into a method for readability

lib/drives/pages/drive_detail_page.dart
fynsta: Yes but for the first one it should say "Start of the Drive" and for the last one "End of the Drive"

fynsta:
By the way, quick question: Do we also need to localize screen reader texts? I did not see a language option in the settings, but it's probably using system default? Which is weird because it's different from the language in the app.

Eddie-42:
> By the way, quick question: Do we also need to localize screen reader texts? I did not see a language option in the settings, but it's probably using system default? Which is weird because it's different from the language in the app.

Yes the screen reader uses the system default and I don't think we can change that. This could lead to problems if the system default differs from the app language as the pronunciation is always messed up really badly, but IMO this is okay because afaik we use the system default as our default langauge, so there are not going to be any issues if the user can speak their default language and I think no app can deal with this properly

fynsta:
Okay then can you localize all those strings? Because this is parallel to the translation PR

istzen:
The time is at the moment just read as numbers. Do we maybe want to change that so that it reads as a time? or is it standard not to do that?

istzen:
When you login, while loading the screen reader starts to read the page from the beginning is there some way to prevent that?

lib/util/trip/trip_overview.dart
```diff
@@ -81,7 +85,12 @@ class TripOverview extends StatelessWidget {
...
       );
-      text = Text("${maxUsedSeats ?? '?'}/${trip.seats} ${S.of(context).seats}");
+      text = Text(
+        "${maxUsedSeats ?? '?'}/${trip.seats} ${S.of(context).seats}",
+        semanticsLabel: maxUsedSeats != null
+            ? S.of(context).labelXOfYseats(maxUsedSeats, trip.seats)
+            : S.of(context).labelUnknownSeats(trip.seats),
+      );
     } else {
       seatIcons = List.generate(
```
istzen, line 95: can we add here that labelXofYSeats also gets read when tapping on the icons or would that be confusing?

Eddie-42:
> The time is at the moment just read as numbers. Do we maybe want to change that so that it reads as a time? or is it standard not to do that?

Yeah I don't know why this happened to you but I can't replicate that error.... Are you okay with me just assuming your screen reader language was different from the one on screen and saying that somehow caused that and we can't do anything about that? :)

lib/util/trip/trip_overview.dart
Eddie-42: Yes I did that, that's much better

MarkFalke:
When the screen reader reads the seats in German in the drive-info it says:"seats plural ist gleich eins other null von eins Sitzen"(I picked 1 possible seat in that case). The ride-info is working correctly.

istzen:
> > When you login, while loading the screen reader starts to read the page from the beginning is there some way to prevent that?
> 
> I know, but I lack the imagination right now to think of a possible solution. The problem is that the transition to the other page is a bit janky because it's done by our AuthApp I think... I would like to push that back on my priority list and maybe fix that in another PR

yes that's okay

istzen:
> > The time is at the moment just read as numbers. Do we maybe want to change that so that it reads as a time? or is it standard not to do that?
> 
> Yeah I don't know why this happened to you but I can't replicate that error.... Are you okay with me just assuming your screen reader language was different from the one on screen and saying that somehow caused that and we can't do anything about that? :)

yes i think that was my fault

istzen (approved):

Showcas (approved):

fynsta (approved):

MarkFalke (approved):

PR #36, opened 2022-12-31
Tests

MarkFalke:
I modified the Code so we can Test it.
To test methods which communicate indirectly with Supabase you have to mock the Supabase method.  
In order for that to work, you have to setup the `when` Method like in the File "accountTest" (but with the method and parameters you need). 
When the Method tries to get the Value of the Database it gets the value you want it to get. So your test doesn't rely on the Data of the Database.
The Methods which doesn't communicate with Supabase at all can be tested normally.

Also i had to refactor all code that communicates with the database into SupabaseManager in order for the Mock to work.


lib/account/models/profile.dart
```diff
@@ -32,7 +31,8 @@ class Profile extends Model {
   });
 
   @override
-  factory Profile.fromJson(Map<String, dynamic> json) {
+  factory Profile.fromJson(Map<String, dynamic> json, ) {
```
fynsta, line 34: what happened here?

lib/account/models/profile.dart
```diff
@@ -52,8 +52,8 @@ class Profile extends Model {
     );
   }
 
-  static List<Profile> fromJsonList(List<Map<String, dynamic>> jsonList) {
```
fynsta, line 55: this definitely needs to stay static, how would that even work non-statically?

lib/account/models/profile.dart
```diff
@@ -77,8 +77,8 @@ class Profile extends Model {
     return 'Profile{id: $id, username: $username, email: $email, createdAt: $createdAt}';
   }
 
-  static Future<Profile?> getProfileFromAuthId(String authId) async {
-    Map<String, dynamic>? query = await supabaseClient.from('profiles').select().eq('auth_id', authId).maybeSingle();
+  static Future<Profile?> getProfileFromAuthId(String authId, SupabaseManager supabaseManager) async {
```
fynsta, line 80: why is the SupabaseManager an argument here?

lib/account/pages/write_review_page.dart
```diff
@@ -36,19 +36,14 @@ class _WriteReviewPageState extends State<WriteReviewPage> {
...
-        .maybeSingle();
+    Map<String, dynamic>? data = await SupabaseManager().getOfDatabaseMapAndLimitMaybeNull(
+      'reviews', '*', 'receiver_id', widget.profile.id, 'writer_id', SupabaseManager().getCurrentProfileMock()!.id!, 1);
     setState(() {
       _review = data != null
           ? Review.fromJson(data)
           : Review(
               rating: 0,
-              writerId: SupabaseManager.getCurrentProfile()!.id!,
+              writerId: SupabaseManager().getCurrentProfileMock()!.id!,
```
fynsta, line 46: mock?

lib/main_app.dart
```diff
@@ -37,8 +33,24 @@ class _MainAppState extends State<MainApp> {
     }
   }
 
+  late Map<TabItem, Widget> _pages;
+
+  @override
+  void initState() {
+    super.initState();
+    _pages = {
```
fynsta, line 41: why did you do this? this should all be constant

lib/util/supabase.dart
```diff
@@ -26,4 +36,68 @@ class SupabaseManager {
       _currentProfile = null;
     }
   }
+
+
```
fynsta, line 40: I'm sorry, WHAT?

- what if I want to use three "eq" operators? Or any operator not used here (there are a lot of filter operators, for example)? Do I need to write a new method for any combination of operators I want to use in any place? We can't do that.
- the arguments here could easily be named hash arguments like we use everywhere. Otherwise you have no clue what to put where when you call the method.
- all of these methods should be static, why are they instance methods?
- "getOf" -> "getFrom" and maybe "selectFrom" is better anyways

I understand you needed to write these methods for mocking, but I'm sure there is a better way.

test/accountTest.dart
```diff
@@ -0,0 +1,20 @@
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:flutter_app/util/supabase.mocks.dart';
+
+void main() async {
+
+MockSupabaseManager mockSupabaseManager = MockSupabaseManager();
```
fynsta, line 8: could you provide some tests so we can see this in action?

fynsta:
I don't get why there are so many whitespacing and formatting changes?
Sometimes you removed "const" despite it needing to be a const constructor.

VSCode shows me 30 Problems
image

fynsta:
the lint is currently erroring, so I assume something is wrong with the syntax in the tests

lib/util/supabase.dart
```diff
@@ -1,5 +1,4 @@
 import 'package:supabase_flutter/supabase_flutter.dart';
```
fynsta, line 1: can you also remove the changes in this file and drive_card?

test/driveTest.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'dart:io';
+import 'package:flutter_app/util/supabase.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'mockServer.dart';
+
+// DIe Klasse UrlProccessor muss zu begin jeder Testdatei implementiert werden und die Methode processUrl überschrieben werden
+// Wird die Methode ProcessUrl aufgrufen, wird für den dort definierten Fall (in dem Beispiel client.from('drives').select('driver_id,seats')) die Antwort definiert
+// Die Datenbankabfrage an den Client wird so abgefangen und das gefünschte ergebnis zurückgegeben.
+// Um herauszifinden welche URL durch die jeweilige Datenbankabfrage generiert wird, einfach den auskommentierten Print-Aufruf in der mockServer.Dart Datei aktivieren
+class DriveProccesor extends UrlProccessor {
```
fynsta, line 11: what if I want different test cases with different responses for the same request ? would I need to write a processor for each one or is there a simpler way?

istzen: What you could do is Mock the method i guess that's probably for most cases easier

istzen (approved):

test/driveTest.dart
fynsta: I don't understand, what exactly would that look like?

test/drive_test.dart
```diff
@@ -0,0 +1,48 @@
+import 'dart:convert';
+import 'dart:io';
+import 'package:motis_mitfahr_app/util/supabase.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'mock_server.dart';
+
+// DIe Klasse UrlProccessor muss zu begin jeder Testdatei implementiert werden und die Methode processUrl überschrieben werden
+// Wird die Methode ProcessUrl aufgrufen, wird für den dort definierten Fall (in dem Beispiel client.from('drives').select('driver_id,seats')) die Antwort definiert
+// Die Datenbankabfrage an den Client wird so abgefangen und das gefünschte ergebnis zurückgegeben.
```
Eddie-42, line 9:
```suggestion
// Die Datenbankabfrage an den Client wird so abgefangen und das gewünschte Ergebnis zurückgegeben.
```

test/mock_server.dart
```diff
@@ -0,0 +1,64 @@
...
+        final foundApiKey = headers.value('apikey');
+        expect(foundApiKey, apiKey);
+        if (foundApiKey == customApiKey) {
+          expect(headers.value('customfield'), 'customvalue');
+        }
+
+        if (expectedFilter != null) {
+          expect(url.contains(expectedFilter), isTrue);
+        }
+        proccessor.processUrl(url, request);
```
Eddie-42, line 34:
```suggestion
        processor.processUrl(url, request);
```

Eddie-42 (changes requested):
Valuable commentary here

lib/home/pages/home-screen.dart
```diff
@@ -0,0 +1,41 @@
+// import 'package:flutter/material.dart';
```
fynsta, line 1: what is this file for?

fynsta:
fyi I added the testing build step back to the pipeline. But for some reason, the whole pipeline is just not running for this branch. any idea why?

lib/home/pages/home-screen.dart
MarkFalke: it was there on accident.

test/drive_test.dart
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  group('basic test', () {
+    test('test with explicit URL', () async {
+      when(driveProcesor.processUrl('/rest/v1/drives?select=driver_id%2Cseats&driver_id=eq.1')).thenReturn(jsonEncode([
+        {'drive_id': 1, 'seats': 2},
+        {'drive_id': 2, 'seats': 1},
+      ]));
+      final data = await SupabaseManager.supabaseClient.from('drives').select('driver_id, seats').eq('driver_id', '1');
+      print(data);
```
fynsta, line 34: can you also add the expectations so the build job has something to run? :)

test/drive_test.dart
```diff
@@ -0,0 +1,54 @@
...
+// Die Datenbankabfrage an den Client wird so abgefangen und das gewünschte Ergebnis zurückgegeben.
+// Um herauszufinden welche URL durch die jeweilige Datenbankabfrage generiert wird, einfach den auskommentierten Print-Aufruf in der mockServer.Dart Datei aktivieren
+
+void main() {
+  MockUrlProcessor driveProcesor = MockUrlProcessor();
+  //setup muss in jeder Testklasse einmal aufgerufen werden
+  setUp(() async {
+    await MockServer.initialize();
+    MockServer.handleRequests(driveProcesor);
+    SupabaseManager.setClient(MockServer.client);
```
Eddie-42, line 20: Can you put this in the MockServer.initialize method?

test/mock_server.dart
```diff
@@ -0,0 +1,62 @@
...
+  static const customApiKey = 'customApiKey';
+  static const apiKey = 'supabaseKey';
+
+  static late SupabaseClient client;
+  static late HttpServer mockServer;
+  static final Set<String> listeners = {};
+  static WebSocket? webSocket;
+  static StreamSubscription<dynamic>? listener;
+
+  static Future<void> handleRequests(UrlProcessor processor, {String? expectedFilter}) async {
```
Eddie-42, line 17: What does expectedFilter do?

test/drive_test.dart
```diff
@@ -0,0 +1,54 @@
...
+    test('test with explicit URL', () async {
+      when(driveProcesor.processUrl('/rest/v1/drives?select=driver_id%2Cseats&driver_id=eq.1')).thenReturn(jsonEncode([
+        {'drive_id': 1, 'seats': 2},
+        {'drive_id': 2, 'seats': 1},
+      ]));
+      final data = await SupabaseManager.supabaseClient.from('drives').select('driver_id, seats').eq('driver_id', '1');
+      print(data);
+    });
+    test('test without explicit URL', () async {
+      when(driveProcesor.processUrl(any)).thenReturn(jsonEncode([
```
Eddie-42, line 37:
```suggestion
      when(driveProcessor.processUrl(any)).thenReturn(jsonEncode([
```

test/drive_test.dart
```diff
@@ -0,0 +1,54 @@
...
+    SupabaseManager.setClient(MockServer.client);
+  });
+  //tearDown muss in jeder Testklasse einmal aufgerufen werden
+  tearDown(() async {
+    MockServer.tearDown();
+  });
+
+  group('basic test', () {
+    test('test with explicit URL', () async {
+      when(driveProcesor.processUrl('/rest/v1/drives?select=driver_id%2Cseats&driver_id=eq.1')).thenReturn(jsonEncode([
```
Eddie-42, line 29:
```suggestion
      when(driveProcessor.processUrl('/rest/v1/drives?select=driver_id%2Cseats&driver_id=eq.1')).thenReturn(jsonEncode([
```

test/drive_test.dart
istzen: good idea

test/mock_server.dart
```diff
@@ -0,0 +1,62 @@
...
+    client = SupabaseClient(
+      'http://${mockServer.address.host}:${mockServer.port}',
+      apiKey,
+      headers: {
+        'X-Client-Info': 'supabase-flutter/0.0.0',
+      },
+    );
+  }
+
+  static void tearDown() async {
```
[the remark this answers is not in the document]
fynsta: Well now the method does not do anything anymore. What is going on? Maybe the listeners are actually needed for something, but we should really know what that is.

istzen:
the idea for the MockServer comes from [here](https://github.com/supabase/supabase-dart/blob/main/test/mock_test.dart)

Eddie-42 (approved):

PR #39, opened 2023-01-05
Add review page

fynsta:
Written by @Eddie-42 

Contains
* A reviews page, where reviews of a profile are shown. You can also rate a user here if you have had a ride with him as driver in the past.
* Refactorings: 

  * AggregateReviewWidget, which contains the overall rating and category ratings. This is used in the Ride detail page and reviews page and therefore deserves its own class.
  * ProfileWidget now includes an InkWell that will lead you to the profile page in the future. To accomplish this, we refactored some places where the InkWell encompassed other elements as well, and provided the option to include all of that in the ProfileWidget.


lib/account/pages/reviews_page.dart
```diff
@@ -0,0 +1,122 @@
...
+      )''').eq('id', _profileId).single();
+    List<dynamic> commonRidesData = await supabaseClient.from('rides').select('''
+          *,
+          drive: drives!inner(*)
+        ''').eq('rider_id', SupabaseManager.getCurrentProfile()!.id).eq('drive.driver_id', _profileId);
+
+    setState(() {
+      _profile = Profile.fromJson(profileData);
+      List<Ride> rides = Ride.fromJsonList(commonRidesData);
+      _reviewable = rides.any((ride) => ride.isFinished && ride.status == RideStatus.approved);
```
fynsta, line 57: this should also check if there is an existing review and change the button then?

lib/account/pages/reviews_page.dart
```diff
@@ -0,0 +1,122 @@
...
+
+import '../../util/review_detail.dart';
+import '../models/review.dart';
+import 'write_review_page.dart';
+
+class ReviewsPage extends StatefulWidget {
+  final int profileId;
+  final Profile? profile;
+
+  const ReviewsPage({super.key, required this.profileId}) : profile = null;
```
istzen, line 18: Do we really need this? because there should not be any calls to the reviewPage without a Profile or am I missing something?

lib/account/pages/reviews_page.dart
istzen: that would probably be a good idea otherwise that might be confusing if I'm riding with a person again and don't remember...

lib/account/pages/reviews_page.dart
Showcas: Its necessary for the anonymous search. Since they can look at the Ratings by going from SearchCard to the Ride Detail and then to the Rating.

Showcas (approved):

lib/account/pages/reviews_page.dart
istzen: shouldn't there the Profile of the Person who was Reviewed been loaded?

istzen:
Why has one check not completed?

lib/account/pages/reviews_page.dart
fynsta: @istzen is probably right, but this is just to be consistent with the other pages and to maybe support that in the future

fynsta:
> Why has one check not completed?

No idea, probably a Github bug

Showcas:
> 



> > Why has one check not completed?
> 
> No idea, probably a Github bug

I just let it rerun and it worked

lib/account/pages/reviews_page.dart
Eddie-42: Done.

MarkFalke (approved):

PR #40, opened 2023-01-05
Profile page

fynsta:
co-authored by @Eddie-42 
description written by @Eddie-42 

Hey guys :) We have some little changes (like 1700) for you :))

First and foremost, the **profile page**! This page shows all information we have about profiles, and if you visit your own page, you can edit everything except your email. 

* We are especially proud of how you can edit your profile features, which is a neat solution to the ordering problem in cards. Note how you can't select two mutually exclusive features, such as _smoking_ and _no smoking_.
* You can upload profile pictures from your gallery. In order to do this, we have a new library: ImagePicker. Therefore, for iOS, this PR includes some files that are needed for permission handling.
* If you click on the avatar, it shows it bigger on a separate screen, as you would expect.
* If you are viewing a profile other than yourself, you can report it while giving a reason. This is stored in the database (please also review the supabase table and RLS policies. Users should only be permitted to insert and read their own reports!), and you cannot review the same person again for 3 days.

Also, a few refactorings, nothing big. I sincerely urge you to completely disregard how we noticed only now that the title of the ride detail page was translated completely incorrectly. You didn't see that right?

A11y and l10n are coming, but please review the PR already. You can trust us that we aren't going to mess up translations, right?


MarkFalke:
In my opinion there should be another possibility to get to the profile page then just click on account and then on your name.
I think its not that intuitive. Maybe if you click on account you get tu the profile-page and have a button in the right corner to get to the settings. In my opinion people will want to get to their profile more often than to their settings.

istzen:
if i select a profile picture shouldn't my Avatar then also display the Picture?

lib/account/models/profile_feature.dart
```diff
@@ -137,4 +143,31 @@ extension FeatureExtension on Feature {
...
+      case Feature.petsAllowed:
+        return other == Feature.noPetsAllowed;
+      case Feature.noChildrenAllowed:
+        return other == Feature.childrenAllowed;
+      case Feature.childrenAllowed:
+        return other == Feature.noChildrenAllowed;
+      case Feature.relaxedDrivingStyle:
+        return other == Feature.speedyDrivingStyle;
+      case Feature.speedyDrivingStyle:
+        return other == Feature.relaxedDrivingStyle;
```
istzen, line 168: Maybe also talkative and quiet?

lib/account/pages/edit_account/edit_birth_date_page.dart
```diff
@@ -0,0 +1,94 @@
...
+      onPressed: () {
+        _controller.clear();
+        _date = null;
+      },
+    );
+  }
+
+  void _showDatePicker() async {
+    final twelveYearsAgo = DateTime.now().subtract(const Duration(days: 365 * 12));
+    DateTime? date = await showDatePicker(
```
istzen, line 75: can we use another DatePicker here or that you can just type the date. I think it's a little bit inconvenient since you have to change the Year and Month of the DatePicker. We could for Example use de Cupertino DatePicker there you can just scroll through the Years, Month but not sure if that works for Android

istzen:
> if i select a profile picture shouldn't my Avatar then also display the Picture?
nevermind my mistake. It does that but in the Account Page it doesn't reload when i change the picture that should be changed.

lib/account/pages/edit_account/edit_birth_date_page.dart
```diff
@@ -0,0 +1,94 @@
...
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(S.of(context).pageProfileEditBirthDateTitle),
+      ),
+      body: Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 10),
+        child: Column(
+          children: [
+            TextField(
```
istzen, line 38: if we have the datepicker this should be read-only i guess

lib/account/models/profile_feature.dart
Eddie-42: Yes please! I also argued with Fynn about that

lib/account/pages/edit_account/edit_birth_date_page.dart
Eddie-42: There is actually a button within the Date picker that lets you type it in. I think we should always use the picker of the target platform, as an Android user I would find an iOS picker really weird. (To be fair, there are far better pickers used in a lot of Android Apps, I would argue in favor of using a package to replace the current picker in another PR)

lib/account/models/profile_feature.dart
fynsta: Lol yeah I know but we agreed to only forbid exact opposites. We are not the profile feature police. If someone wants to talk half the journey and be quiet the other half, who are we to stand in the way of that?

lib/account/pages/edit_account/edit_birth_date_page.dart
fynsta: yes, good point 👍

Eddie-42:
> In my opinion there should be another possibility to get to the profile page then just click on account and then on your name. I think its not that intuitive. Maybe if you click on account you get tu the profile-page and have a button in the right corner to get to the settings. In my opinion people will want to get to their profile more often than to their settings.

Let's talk about that on Monday, we were inspired by the design of Whatsapp or Signal. I think if we want to keep the current structure, it would help to rename the account tab to settings and display an edit icon within the profile Widget. I would personally prefer that solution, but we can talk about that or just wait until the user survey

fynsta:
>  In my opinion there should be another possibility to get to the profile page then just click on account and then on your name.
I think its not that intuitive. Maybe if you click on account you get tu the profile-page and have a button in the right corner to get to the settings. In my opinion people will want to get to their profile more often than to their settings.

That's interesting, we should probably talk about that in the next meeting. I might agree with you, but I'll have to think about it. (I also got confused by the distinction between account and profile page all the time, as Simon could tell you)

fynsta:
> > if i select a profile picture shouldn't my Avatar then also display the Picture?

> nevermind my mistake. It does that but in the Account Page it doesn't reload when i change the picture that should be changed.

That's correct, but I'm not sure how to fix it. Those pages would have to be notified that the underlying data changed. Currently our whole system is not set up that way. Maybe it makes sense to add the RefreshIndicator to a few more pages so that the user gets notified, but other than that I don't see good solutions.

lib/account/models/profile_feature.dart
Showcas: but if someone does both wouldn't they just not mark any of the features. I understand talkative as if you can, you should talk and quiet means if you can please stay quiet. If you do not perfere one or the other then don't mark it as a feature?

fynsta: But that gets us in edge cases real fast. Should "talkative" and "music" also be exclusive? You can't do both at the same time either

istzen: Maybe Fynn is Right. I mean in the End it is not our problem.

lib/account/pages/edit_account/edit_birth_date_page.dart
istzen: sounds good.

istzen:
> > > if i select a profile picture shouldn't my Avatar then also display the Picture?
> 
> 
> 
> > nevermind my mistake. It does that but in the Account Page it doesn't reload when i change the picture that should be changed.
> 
> 
> 
> That's correct, but I'm not sure how to fix it. Those pages would have to be notified that the underlying data changed. Currently our whole system is not set up that way. Maybe it makes sense to add the RefreshIndicator to a few more pages so that the user gets notified, but other than that I don't see good solutions.

Can't we just reload the Profile when going Back i thought we did that in a few other cases

lib/account/models/profile_feature.dart
Showcas: sure, It does not hurt to leave it in atm.

Showcas:
Im still not convinced with the drag and drop system for the Feature selection.  I'd prefer a + for unused features to be added and a - for remuving them again. You could leave the for the drag and drop to work in the added features so you can declare your priotiries. Just for me when I use it, it is kinda hard to pull a feature from the bottom all the way to the top with scrolling

istzen:
shouldn't there also be the possibility to remove the Profile Picture. Or do i just not realize how this can be done.

fynsta:
> Can't we just reload the Profile when going Back i thought we did that in a few other cases

Yes, but that would not be a complete solution... That would only work for the AccountPage and only if the ProfilePage is visited via that Navigation stack exactly. Also the accountPage is using the profile from the SupabaseManager, not anything that is loaded anyways.

fynsta:
> Im still not convinced with the drag and drop system for the Feature selection. I'd prefer a + for unused features to be added and a - for remuving them again. You could leave the for the drag and drop to work in the added features so you can declare your priotiries. Just for me when I use it, it is kinda hard to pull a feature from the bottom all the way to the top with scrolling

Okay, but that's probably just because of the touch pad. With "+" and "-", there is no way to reorder them. We can maybe look into adding them, but that probably would be a very cluttered interface.

> shouldn't there also be the possibility to remove the Profile Picture. Or do i just not realize how this can be done.

Yeeah good point, that does not exist yet. But I don't even know how the UI for something like that would look.

istzen:
> > Can't we just reload the Profile when going Back i thought we did that in a few other cases
> 
> Yes, but that would not be a complete solution... That would only work for the AccountPage and only if the ProfilePage is visited via that Navigation stack exactly. Also the accountPage is using the profile from the SupabaseManager, not anything that is loaded anyways.

Can we then just reload that Profile? and for the other i think it's not that problematic since there the user does not see it directly, since it's not his own picture

istzen:
> > > Can't we just reload the Profile when going Back i thought we did that in a few other cases
> > 
> > 
> > Yes, but that would not be a complete solution... That would only work for the AccountPage and only if the ProfilePage is visited via that Navigation stack exactly. Also the accountPage is using the profile from the SupabaseManager, not anything that is loaded anyways.
> 
> Can we then just reload that Profile? and for the other i think it's not that problematic since there the user does not see it directly, since it's not his own picture

we can use there something like WhatsApp there you have an Option for deleting the picture.

Showcas:
> h

I guess we can leave it as it is at the moment and the user study will tell us if it is better or not

lib/account/widgets/avatar.dart
```diff
@@ -0,0 +1,125 @@
...
+            right: 0,
+            bottom: 0,
+            child: CircleAvatar(
+              radius: 20,
+              backgroundColor: Theme.of(context).colorScheme.primary,
+              child: IconButton(
+                tooltip: S.of(context).widgetAvatarUploadTooltip,
+                iconSize: 20,
+                onPressed: _isLoading ? null : _upload,
+                icon: const Icon(Icons.camera_alt),
```
Showcas, line 78: Icons.camera_alt as an icon gave me the intuition that it is for taking a picture with the kamera and then to put it as a Avatar.
Icons.photo represents the funktion of picking a picture out of photo better.

fynsta: I agree. There is actually also Icons.photo_library, which is even more appropriate

fynsta:
> we can use there something like WhatsApp there you have an Option for deleting the picture.

Yes sure, but they also a custom dialog for this, which seems like lot more work. We'll maybe have to think about how to prioritize this...

fynsta:
I have now added plus/minus buttons that can additionally be used. Do you think this is better? Should we remove the drag handles? I am not sure what is the best UX here.

lib/account/widgets/avatar.dart
```diff
@@ -0,0 +1,125 @@
...
+                icon: const Icon(Icons.photo_library),
+              ),
+            ),
+          ),
+      ],
+    );
+  }
+
+  Future<void> _upload() async {
+    final picker = ImagePicker();
```
Showcas, line 87: bei mir wirft der hier momentan ein Error Weil ImagePicker nicht deklariert ist

Showcas:
> I have now added plus/minus buttons that can additionally be used. Do you think this is better? Should we remove the drag handles? I am not sure what is the best UX here.

I like it.

Showcas (approved):

istzen:
> > we can use there something like WhatsApp there you have an Option for deleting the picture.
> 
> Yes sure, but they also a have custom dialog for this, which seems like lot more work. We'll maybe have to think about how to prioritize this...

yes but the Dialog shouldn't be much work should it? but you are right this does not have the highest priority.

fynsta:
Okay, automatically reloading on account page works now. The other pages would only reload if you called setState on each one, but I think this is good enough.

istzen:
> Okay, automatically reloading on account page works now. The other pages would only reload if you called setState on each one, but I think this is good enough.

perfect. Did we want to change the accountPage and settingsPage or not? otherwise can you add some kind of indicator to the Profilepage and maybe rename account to settings in the Tabs

istzen:
and i still think it would be nice if one could delete the picture

fynsta:
> perfect. Did we want to change the accountPage and settingsPage or not? otherwise can you add some kind of indicator to the Profilepage and maybe rename account to settings in the Tabs

I think it should still be called Account. We will add a sign on the widget like this next to the name: ">". And see how that goes in the user study.

Deleting the picture will be added.

MarkFalke (approved):

istzen:
> > > perfect. Did we want to change the accountPage and settingsPage or not? otherwise can you add some kind of indicator to the Profilepage and maybe rename account to settings in the Tabs
> > 
> > 
> > I think it should still be called Account. We will add a sign on the widget like this next to the name: ">". And see how that goes in the user study.
> > Deleting the picture will be added.
> 
> Both done.

perfect thanks. We could also add the logout button on the Settings page what do you think?

istzen (approved):

Eddie-42 (approved):

Eddie-42:
> > > > perfect. Did we want to change the accountPage and settingsPage or not? otherwise can you add some kind of indicator to the Profilepage and maybe rename account to settings in the Tabs
> > > 
> > > 
> > > I think it should still be called Account. We will add a sign on the widget like this next to the name: ">". And see how that goes in the user study.
> > > Deleting the picture will be added.
> > 
> > 
> > Both done.
> 
> perfect thanks. We could also add the logout button on the Settings page what do you think?

I wouldn't provide MORE ways to log out :D For such things I think it is best to build in a few Try-to-delete-your-account-on-Amazon-y dark patterns! No seriously, I think since your session is saved, logging in and out isn't really first page core functionality but we can discuss that tomorrow

PR #41, opened 2023-01-08
fix Time-/DatePicker for search suggestion page

istzen:
I fixed the Time-/DatePicker for search suggestion page since they did not work at the moment. 
Thanks @Cedric for noticing.

### DIFF
--- a/lib/rides/pages/search_suggestion_page.dart
+++ b/lib/rides/pages/search_suggestion_page.dart
@@ -64,8 +64,11 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
       lastDate: firstDate.add(const Duration(days: 30)),
     ).then((value) {
       if (value != null) {
-        _selectedDate = DateTime(value.year, value.month, value.day, _selectedDate.hour, _selectedDate.minute);
-        _dateController.text = localeManager.formatDate(_selectedDate);
+        setState(() {
+          _selectedDate = DateTime(value.year, value.month, value.day, _selectedDate.hour, _selectedDate.minute);
+          _dateController.text = localeManager.formatDate(_selectedDate);
+        });
+        if (_timeValidator(_timeController.text) == null) loadRides();
       }
     });
   }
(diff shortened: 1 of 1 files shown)


Showcas (approved):

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -173,37 +180,31 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
 
   Widget buildTimePicker() {
-    _timeController.text = localeManager.formatTime(widget.date);
+    _timeController.text = localeManager.formatTime(_selectedDate);
 
     return TextFormField(
       decoration: InputDecoration(
         border: const OutlineInputBorder(),
         labelText: S.of(context).formTime,
+        errorText: _timeValidator(_timeController.text),
```
fynsta, line 203: why does this work instead of `validator: _timeValidator`?

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -64,8 +64,11 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
       lastDate: firstDate.add(const Duration(days: 30)),
     ).then((value) {
       if (value != null) {
-        _selectedDate = DateTime(value.year, value.month, value.day, _selectedDate.hour, _selectedDate.minute);
-        _dateController.text = localeManager.formatDate(_selectedDate);
+        setState(() {
+          _selectedDate = DateTime(value.year, value.month, value.day, _selectedDate.hour, _selectedDate.minute);
+          _dateController.text = localeManager.formatDate(_selectedDate);
+        });
+        if (_timeValidator(_timeController.text) == null) loadRides();
```
fynsta, line 71: why does only the time need to be validated? And when the time is wrong, shouldn't we show some kind of error message?

lib/rides/pages/search_suggestion_page.dart
istzen: because there is no Form around it so the validator never gets called

lib/rides/pages/search_suggestion_page.dart
istzen: because the date is validated by the picker which only allows to select dates in the future. And there is a error message displayed. Thats what the errorText is for

lib/rides/pages/search_suggestion_page.dart
fynsta: Ah yes, that makes sense! But why isn't there a form on that page in the first place? because it's never submitted?

istzen: yes i think so you have to ask @Showcas for that. But it works like this.

Eddie-42 (approved):

MarkFalke (approved):

PR #42, opened 2023-01-09
Garbled mess

Eddie-42:
I'm sorry, there is no real focus with this PR. It just does some unrelated stuff:
* **Buttons**, mainly... I refactored buttons so that we can merge many files into one... Instead of `BigButton`, `SubmitButton` and `color` as a parameter, there is `Button` (former `BigButton`), `Button.submit` (former `SubmitButton`) and `Button.error` (former `Button(color: Theme.of(context).colorScheme.error`).
* The error that shows when logging in should be gone. It came from the Anonymous Search page because that had both a Hero and a FloatingActionButton. Unbeknownst to most everyone, FloatingActionButtons actually create heros internally without us specifying it and without telling anyone, and so there were multiple heros on that page, which can be problematic. The next problem is that heros just don't work within our normal, non-login tab pages right now. The reason for the FABs including heros out of the box is that it can create a beautiful effect when the FAB gets bigger and animates into the page it opens.
* FAQ cards are animated now. The biggest and most important of changes!


lib/main_app.dart
```diff
@@ -42,8 +42,11 @@ class _MainAppState extends State<MainApp> {
   Widget build(BuildContext context) {
     return WillPopScope(
       onWillPop: () async {
+        print("Deciding");
```
fynsta, line 45: debug prints still here

Showcas:
Looks good to me. So resulting PRs are that we need to fix CreateDrive and add the Email-link in help?
I understood it that heros don't work. Then why are we using them now?

istzen:
if i change from dark to light mode an Exception is thrown.

lib/main_app.dart
Eddie-42: You saw nothing

Eddie-42:
> if i change from dark to light mode an Exception is thrown.



> if i change from dark to light mode an Exception is thrown.

Have you tried restarting the emulator and cold restarting the app and if you have, can you share a screenshot of the exception? I can't reproduce that

istzen (approved):

fynsta (approved):

MarkFalke (approved):

Showcas (approved):

PR #43, opened 2023-01-10
change cancel drive

istzen:
I added a supabase trigger that updates the rides automatically as we cancel the drive. Therefore we don't net the code for that anymore and also the RLS policy that allows drivers to update rides. 
Pleas review the function in Supabase [here](https://app.supabase.com/project/khvzghkhzkjmnbagbzkm/database/functions) and the trigger [here](https://app.supabase.com/project/khvzghkhzkjmnbagbzkm/database/triggers) you can test if it works also via the admin. Just to avoid confusion the new Parameter in the function is the changed row, this parameter is in every trigger available and has to be returned at the end.

### DIFF
--- a/lib/drives/models/drive.dart
+++ b/lib/drives/models/drive.dart
@@ -134,12 +134,7 @@ class Drive extends Trip {
   Future<void> cancel() async {
     cancelled = true;
     await supabaseClient.from('drives').update({'cancelled': true}).eq('id', id);
-    //cancel all pending and approved rides for this drive.
-    await supabaseClient
-        .from('rides')
-        .update({'status': RideStatus.cancelledByDriver.index})
-        .eq('drive_id', id)
-        .or('status.eq.${RideStatus.pending.index},status.eq.${RideStatus.approved.index}');
+    //the rides get updated automatically by a supabase function.
   }
 
   @override


Eddie-42 (approved):

fynsta:
Seems to work well overall! A few notes:

1. Would it be possible for the function to check the old row? So that this actually only happens when the cancelled value is changed
2. While testing this on RailsAdmin, I noticed that the enums there are off by one? It seems to start counting the enum with 1 instead of 0, which leads to some very weird-looking behaviour.

fynsta (approved):

istzen:
> Seems to work well overall! A few notes:
> 
> 1. Would it be possible for the function to check the old row? So that this actually only happens when the cancelled value is changed
> 2. While testing this on ReactAdmin, I noticed that the enums there are off by one? It seems to start counting the enum with 1 instead of 0, which leads to some very weird-looking behaviour.

1. i looked it up and it seems that there is also an old variable that i can use. I'll change it 
2. yes i noticed that too i changed it but i guess i forgot to push then.

fynsta (approved):

Showcas (approved):

MarkFalke (approved):

PR #44, opened 2023-01-10
Add about page

Eddie-42:
An about page! I'm not too sure about the text (most of which would be better suited on a terms of service website). Also, I don't even know what a build signature is , but I'm sure all of you would agree that it looks cool and **VERY** professional (and chances are it could help developers if users experience bugs, I'm sure that's the sort of thing that goes into automated bug reports and since we don't have that, it can just be displayed here I guess).

### DIFF
--- a/lib/account/pages/about_page.dart
+++ b/lib/account/pages/about_page.dart
@@ -4,6 +4,9 @@ import 'package:flutter_svg/svg.dart';
 import 'package:url_launcher/url_launcher_string.dart';
 import 'package:package_info_plus/package_info_plus.dart';
 
+const String githubLink = 'https://github.com/MOTIS-Mitfahr-App/flutter-app';
+const String websiteLink = 'https://motis-project.de/';
+
 class AboutPage extends StatefulWidget {
   const AboutPage({super.key});
 
@@ -26,6 +29,10 @@ class _AboutPageState extends State<AboutPage> {
 
   @override
   Widget build(BuildContext context) {
+    TextStyle packageInfoStyle = Theme.of(context)
+        .textTheme
+        .subtitle2!
+        .copyWith(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5));
     return Scaffold(
       appBar: AppBar(
         title: Text(S.of(context).pageAboutTitle),
@@ -42,31 +49,22 @@ class _AboutPageState extends State<AboutPage> {
                   Expanded(child: Container()),
                   Text(
                     S.of(context).appName,
-                    style: Theme.of(context).primaryTextTheme.displaySmall,
+                    style: Theme.of(context).textTheme.displaySmall,
                   ),
                   if (_packageInfo != null) ...[
                     Text(
                       S.of(context).pageAboutVersion(_packageInfo!.version),
-                      style: Theme.of(context)
-                          .primaryTextTheme
-                          .subtitle2!
-                          .copyWith(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5)),
+                      style: packageInfoStyle,
                       textAlign: TextAlign.center,
                     ),
                     Text(
                       S.of(context).pageAboutBuildNumber(_packageInfo!.buildNumber),
-                      style: Theme.of(context)
-                          .primaryTextTheme
-                          .subtitle2!
-                          .copyWith(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5)),
+                      style: packageInfoStyle,
                       textAlign: TextAlign.center,
                     ),
                     Text(
                       S.of(context).pageAboutBuildSignature(_packageInfo!.buildSignature),
-                      style: Theme.of(context)
-                          .primaryTextTheme
-                          .subtitle2!
-                          .copyWith(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5)),
+                      style: packageInfoStyle,
                       textAlign: TextAlign.center,
                     ),
                   ],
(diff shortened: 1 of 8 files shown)


lib/account/pages/about_page.dart
```diff
@@ -9,15 +12,98 @@ class AboutPage extends StatefulWidget {
...
+                    textAlign: TextAlign.center,
+                  ),
+                  Expanded(
+                    child: Align(
+                      alignment: Alignment.bottomCenter,
+                      child: Row(mainAxisAlignment: MainAxisAlignment.center, children: [
+                        TextButton(
+                          onPressed: () async => await launchUrlString(
+                              'https://github.com/MOTIS-Mitfahr-App/flutter-app',
+                              mode: LaunchMode.externalApplication),
```
fynsta, line 88: comma

fynsta:
Just fyi I pushed compilation for macos/ios

lib/account/pages/about_page.dart
```diff
@@ -9,15 +12,101 @@ class AboutPage extends StatefulWidget {
...
+          slivers: [
+            SliverFillRemaining(
+              hasScrollBody: false,
+              child: Padding(
+                padding: const EdgeInsets.all(10),
+                child: Column(children: [
+                  Expanded(child: Container()),
+                  Text(
+                    S.of(context).appName,
+                    style: Theme.of(context).primaryTextTheme.displaySmall,
```
fynsta, line 45: Is this supposed to look this weird? 
image

This seems to be the wrong style.

lib/account/pages/about_page.dart
```diff
@@ -9,15 +12,101 @@ class AboutPage extends StatefulWidget {
...
+                  ),
+                  if (_packageInfo != null) ...[
+                    Text(
+                      S.of(context).pageAboutVersion(_packageInfo!.version),
+                      style: Theme.of(context)
```
fynsta, line 50: move the style into a variable? it's always the same

lib/account/pages/about_page.dart
```diff
@@ -9,15 +12,101 @@ class AboutPage extends StatefulWidget {
...
+                          .subtitle2!
+                          .copyWith(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5)),
+                      textAlign: TextAlign.center,
+                    ),
+                  ],
+                  SvgPicture.asset(
+                    'assets/logo.svg',
+                    width: MediaQuery.of(context).size.width / 2,
+                    height: MediaQuery.of(context).size.width / 2,
+                    semanticsLabel: 'MOTIS Logo',
```
fynsta, line 77: not internationalized

lib/account/pages/about_page.dart
```diff
@@ -9,15 +12,101 @@ class AboutPage extends StatefulWidget {
...
+                    textAlign: TextAlign.center,
+                  ),
+                  Expanded(
+                    child: Align(
+                      alignment: Alignment.bottomCenter,
+                      child: Row(mainAxisAlignment: MainAxisAlignment.center, children: [
+                        TextButton(
+                          onPressed: () async => await launchUrlString(
+                            'https://github.com/MOTIS-Mitfahr-App/flutter-app',
+                            mode: LaunchMode.externalApplication,
```
fynsta, line 90: why does this have another launch mode, but not the motis links?

lib/account/pages/about_page.dart
```diff
@@ -9,15 +12,101 @@ class AboutPage extends StatefulWidget {
...
+                    S.of(context).pageAboutText,
+                    textAlign: TextAlign.center,
+                  ),
+                  Expanded(
+                    child: Align(
+                      alignment: Alignment.bottomCenter,
+                      child: Row(mainAxisAlignment: MainAxisAlignment.center, children: [
+                        TextButton(
+                          onPressed: () async => await launchUrlString(
+                            'https://github.com/MOTIS-Mitfahr-App/flutter-app',
```
fynsta, line 89: it would be better to move these links into constants at the top of the file/class

lib/account/pages/about_page.dart
Eddie-42: Good point, I only tested this with dark mode. Fixed

lib/account/pages/about_page.dart
Eddie-42: Yes I did that, but I'd like to make a new color scheme entry for this anyways... There are many areas in the code where we use this greyed out onSurface with opacity

lib/account/pages/about_page.dart
Eddie-42: Thanks, done

lib/account/pages/about_page.dart
Eddie-42: The default (at least on Android) is that the link opens a kind of weaker in-app browser without tabs or anything that also doesn't cache where it was when the app is closed. I really like that for the website since you don't get the feeling you are leaving the app, but I thought if you want to browse the GitHub project you will want to have it open in your normal browser, especially so that you can keep browsing after the app is closed

lib/account/pages/about_page.dart
fynsta: Nice, thanks! I stillthink this is maybe too big? In english, it already flows into the second line on the device I'm testing on

Eddie-42: Better?

fynsta (approved):

Showcas (approved):

lib/account/pages/about_page.dart
```diff
@@ -1,5 +1,11 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+import 'package:flutter_svg/svg.dart';
+import 'package:url_launcher/url_launcher_string.dart';
+import 'package:package_info_plus/package_info_plus.dart';
+
+const String githubLink = 'https://github.com/MOTIS-Mitfahr-App/flutter-app';
+const String websiteLink = 'https://motis-project.de/';
```
istzen, line 8: i know we don't have a terms of service link but can we add that just for completeness and use the same as websiteLink for now.

MarkFalke (approved):

istzen (approved):

PR #46, opened 2023-01-11
Store addresses in database

fynsta:
@istzen here it is!!!!!

Now we finally actually store the address suggestion (name + position) in the database. This will not change anything on the surface, but would allow the algorithm to work correctly.

Note the ".toDouble" when getting the values from the database... dart is not smart enough to get a number via json and then cast it to double automatically.

### DIFF
--- a/lib/rides/pages/search_ride_page.dart
+++ b/lib/rides/pages/search_ride_page.dart
@@ -5,6 +5,8 @@ import 'package:motis_mitfahr_app/util/search/address_search_field.dart';
 import 'package:motis_mitfahr_app/util/submit_button.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 
+import '../../util/search/address_suggestion.dart';
+
 class SearchRidePage extends StatefulWidget {
   const SearchRidePage({super.key});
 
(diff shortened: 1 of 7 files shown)


Eddie-42 (approved):

lib/rides/pages/search_ride_page.dart
```diff
@@ -196,12 +204,16 @@ class _SearchRideFormState extends State<SearchRideForm> {
           children: [
             AddressSearchField.start(
               controller: _startController,
-              onSelected: (suggestion) {},
+              onSelected: (suggestion) {
+                _startSuggestion = suggestion;
```
Showcas, line 208: Can you add read only? So we can fix the bug with being able to write without the Addressuggestion

Showcas: nevermind apparently readOnly does not solve our issue. Since we tried that alread in CreateDrive

lib/util/search/address_search_field.dart
```diff
@@ -44,6 +44,7 @@ class AddressSearchField extends StatelessWidget {
       label: getLabelText(context),
       tooltip: getHintText(context),
       excludeSemantics: true,
+      readOnly: true,
```
Showcas, line 47: Add the read only to the TextFormFile and then it works

MarkFalke (approved):

PR #48, opened 2023-01-11
Withdraw

Eddie-42:
Enables users to withdraw their ride request. In order to do this, I added a new RideStatus (and fixed a few bugs concening which information is shown with which RideStatus along the way).

Also, I cleaned up some problems that have been solved in the previous PRs.


lib/util/profiles/profile_widget.dart
```diff
@@ -20,6 +21,7 @@ class ProfileWidget extends StatelessWidget {
     this.showDescription = false,
     this.actionWidget,
     this.isTappable = true,
+    this.customAction,
     this.onPop,
```
fynsta, line 25: 1) customAction should be called "onTap"
2) please make it a Function() instead of VoidCallback
3) there should be an assertion that onTap and onPop cannot be set at the same time

fynsta (approved):

istzen:
the assertions are not fullfilled in the AccountPage

istzen:
can we change the color of the withdraw button so it looks different from the request button?

lib/rides/models/ride.dart
```diff
@@ -147,13 +147,18 @@ class Ride extends Trip {
...
+  }
+
   @override
   String toString() {
     return 'Ride{id: $id, in: $driveId, from: $start at $startTime, to: $end at $endTime, by: $riderId}';
   }
 }
 
-enum RideStatus { preview, pending, approved, rejected, cancelledByDriver, cancelledByRider }
+enum RideStatus { preview, pending, approved, rejected, cancelledByDriver, cancelledByRider, withdrawnByRider }
```
istzen, line 161: what we could also do if we implement the soft delete that we just delete the ride, and remove the withdrawnByRider status. But i'm not sure if that's a good idea, what do you guys think?

istzen:
i also changed the trigger function so that it works with the new Status

lib/rides/models/ride.dart
Eddie-42: I think having the status is good. It provides us with information we need. An example: I discovered that if you request a ride, withdraw it, go back, tap on the ride again, and repeat the same thing over and over, you can generate a lot a database entries and it would be good to be able to see that this is because of withdrawn rides

Eddie-42:
> the assertions are not fullfilled in the AccountPage

Yes thank you, that was a mistake and it's fixed now

lib/rides/models/ride.dart
fynsta: What we could do here is actually using a hard delete, but it is a question whether we want/need the information that there formerly was a request.

Showcas: I find the hard delete attractive but its a risk if we let the users delete rides in our database. I think we should keep the status, its better for the cases and its not intuitive that withdrawn rides are inside the hide bool.

Showcas:
> > can we change the color of the withdraw button so it looks different from the request button?
> 
> What would you suggest? Secondary? Warning? Do the others think that is even a good idea?

Wouldn't error be the most fitting? Like you are "deleting" something so I would expect red

Showcas (approved):

MarkFalke (approved):

Eddie-42:
Sure, done.

istzen (approved):
sorry for approving so late. I forgot a hadn't done that already

PR #49, opened 2023-01-12
Filter

Eddie-42:
**Filter**!
You can filter 
* ratings, 
* features, 
* the difference between the start time of the ride and the time you searched for, 
* and sort by the aforementioned time difference, price, or relevance (a combination of the two)!

I also redesigned the top of the SearchSuggestionsPage a bit, since the filter button was an eyesore before. Let me know your feedback! (The row below the start and destination timeline is copied from the search ride and create drive pages and can be improved on when their time comes).


fynsta:
what does the shrug asset do? and the arb files should probably not be completely different?

Eddie-42:
You can easily test what the shrug asset does by trying to search for a ride from Poing to Titz!
Yeah I suspect the arb files are due to line endings with Windows, I'll fix that

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -309,19 +348,275 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
+          ),
+        ],
+      ),
+    );
+  }
+
+  Widget _buildDeviationFilter() {
+    return filterCategory(
+      S.of(context).pageSearchSuggestionsFilterDeviation,
+      Row(
```
fynsta, line 500: should also include the ability to just select "whole day"

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -309,19 +348,275 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
+        return S.of(context).pageSearchSuggestionsSortingRelevance;
+      case SearchSuggestionSorting.time:
+        return S.of(context).pageSearchSuggestionsSortingTime;
+      case SearchSuggestionSorting.price:
+        return S.of(context).pageSearchSuggestionsSortingPrice;
+    }
+  }
+
+  int Function(Ride, Ride) sortFunction(DateTime date) {
+    timeFunc(Ride ride1, Ride ride2) => date.difference(ride1.startTime).compareTo(date.difference(ride2.startTime));
```
fynsta, line 611: this should probably just use the difference (maybe scaled a little bit) because we don't really know the scale of compareTo and how it compares to the price

same for the priceFunc

lib/l10n/app_de.arb
```diff
@@ -228,7 +231,19 @@
   "pageSearchRideButtonSearch": "Suchen",
 
   "pageSearchSuggestionsTitle": "Vorschläge",
-  "pageSearchSuggestionsButtonFilter": "Filter",
+  "pageSearchSuggestionsEmpty": "Nichts zu sehen",
+  "pageSearchSuggestionsTooltipFilter": "Filter",
+  "pageSearchSuggestionsFilterMinimumRating": "Mindeste Bewertung",
```
fynsta, line 236: Mindest-Bewertung is better

Showcas:
When I try to select exclusive features the alert is behind the overlay so you don't see it

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -309,19 +349,274 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
+            onPressed: () => innerSetState(() => _isRatingExpanded = !_isRatingExpanded),
+            child: Text(_isRatingExpanded ? S.of(context).retract : S.of(context).expand),
+          ),
+        ],
+      ),
+    );
+  }
+
+  Widget _buildFeaturesFilter(void Function(void Function()) innerSetState) {
+    List<Feature> shownFeatures = _isFeatureListExpanded ? Feature.values : _commonFeatures;
```
fynsta, line 447: this list should be updated when I have chosen a feature. So that I can see my choices when I retract the list again.

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -309,19 +349,274 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
+                _selectedDate.difference(ride.startTime) < Duration(hours: int.parse(_maxDeviationController.text));
+            return ratingSatisfied && featuresSatisfied && maxDeviationSatisfied;
+          },
+        )
+        .sorted(_sorting.sortFunction(_selectedDate))
+        .toList();
+  }
+}
+
+enum SearchSuggestionSorting { relevance, time, price }
```
Showcas, line 596: I think price is good. I would change time to traveltime. And relevance is the price and traveltime together?

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -248,59 +270,77 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
+        }
+      },
+      child: Align(
+        alignment: Alignment.centerLeft,
+        child: Text(controller.text),
       ),
     );
   }
 
   Widget buildSearchCardList() {
```
fynsta, line 297: should we show the applied filters somewhere above this list? (ofc only if they're different from the default)

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -248,59 +270,77 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
+    if (filteredSuggestions.isEmpty) {
+      list = Center(
+        child: SingleChildScrollView(
+          physics: const AlwaysScrollableScrollPhysics(),
+          child: Column(
+            children: [
+              Image.asset('assets/shrug.png'),
+              Text(
+                S.of(context).pageSearchSuggestionsEmpty,
+                style: Theme.of(context).textTheme.displaySmall,
```
fynsta, line 312: this is a little bit too big for my taste. Also, this should probably also link to the filter page if filters are set: "Relax your search restrictions..."

lib/rides/pages/search_suggestion_page.dart
Eddie-42: Food for thought: Shouldn't that be an option when looking for a ride? Instead of entering a time, you could check a checkbox "whole day" and then the filter option becomes the max deviation of days

lib/rides/pages/search_suggestion_page.dart
Eddie-42: I've also thought about that, and you have inspired me to implement that. I realized that it would be a crime to not to include a "filter" button in the row that shows you all your filters, and that at least the sorting should probably always be displayed. So, since there is now a filter button in the row and the row is always present, I removed the button from the app bar. In general, this is again a pretty big change; and for this and the whole design I would very much appreciate your feedback :)

Eddie-42:
FYI: I refactored the filters into an own class because the SearchSuggestionPage became a bit messy. This should not affect functionality.

lib/rides/pages/search_suggestion_page.dart
Eddie-42: Done, now 1€ should be equal to 100 minutes... Time is probably valued too high, but I don't know what prices we're dealing with here. What would you suggest?

lib/rides/pages/search_suggestion_page.dart
Eddie-42: I didn't include an outright filter button because there is one probably less than 100 pixels above the text. But this is also very useful if you know that if the search results are really empty (e.g. there is no ride found for that route at all), it displays that too, now

lib/l10n/app_de.arb
```diff
@@ -228,7 +231,22 @@
   "pageSearchRideButtonSearch": "Suchen",
 
   "pageSearchSuggestionsTitle": "Vorschläge",
-  "pageSearchSuggestionsButtonFilter": "Filter",
+  "pageSearchSuggestionsEmpty": "Nichts zu sehen",
+  "pageSearchSuggestionsRelaxRestrictions": "Lockern Sie Ihre Suchbeschränkungen.",
```
fynsta, line 235: duzen :)  und vielleicht lieber noch ein "bitte"

lib/rides/pages/search_suggestion_page.dart
fynsta: Sure we can do it that way too... Is there space for that?

lib/rides/pages/search_suggestion_page.dart
fynsta: I think that's fine, normally you want something pretty close to your search terms

lib/rides/pages/search_suggestion_page.dart
fynsta: Well we can just make the text clickable without making it too obvious? It's just something the user might try out...

lib/rides/pages/search_suggestion_page.dart
fynsta: Oh that looks really interesting! I had just envisioned some blue text that says the filters in words, but this is pretty cool. For the rating categories, I think there should be a bit more separation between the different entries to make it more clear.
And I think the whole row should be clickable to open the filter dialog!

lib/rides/pages/search_suggestion_page.dart
fynsta: Thanks for implementing that! I actually think it makes sense to then only show the chosen features, because the list might get really long (And it's confusing if you show both a chosen feature and its not-chosen mutually exclusive feature).

Another thing: When expanding, the entries should stay in their position (at least the chosen ones). Otherwise the animation with the blue background looks weird.

lib/l10n/app_en.arb
```diff
@@ -279,7 +282,22 @@
   "pageSearchRideButtonSearch": "Search",
 
   "pageSearchSuggestionsTitle": "Search Suggestions",
-  "pageSearchSuggestionsButtonFilter": "Filter",
+  "pageSearchSuggestionsEmpty": "Nothing to see",
```
fynsta, line 285: "Nothing to see _here_"

lib/rides/pages/search_suggestion_page.dart
fynsta: Btw if we do that, we can think about removing the time frame that the user can choose? I know that Blablacar always shows the whole day, and then you can go left and right to view other dates.

lib/rides/widgets/search_suggestion_filter.dart
```diff
@@ -0,0 +1,439 @@
...
+      if (icon != null) ...[icon, const SizedBox(width: 3)],
+      Text(rating.toString()),
+      const Icon(
+        Icons.star,
+        color: Colors.amber,
+      )
+    ]);
+  }
+
+  Widget getRow(BuildContext context, void Function(void Function()) setState) {
```
fynsta, line 275: getIndicatorRow

fynsta: or actually "buildIndicatorRow"

lib/rides/widgets/search_suggestion_filter.dart
```diff
@@ -0,0 +1,439 @@
...
+        return S.of(context).searchSuggestionsSortingRelevance;
+      case SearchSuggestionSorting.time:
+        return S.of(context).searchSuggestionsSortingTime;
+      case SearchSuggestionSorting.price:
+        return S.of(context).searchSuggestionsSortingPrice;
+    }
+  }
+
+  int Function(Ride, Ride) sortFunction(DateTime date) {
+    timeFunc(Ride ride1, Ride ride2) => (date.difference(ride1.startTime) - date.difference(ride2.startTime)).inMinutes;
```
fynsta, line 428: we could declare those inside the extension so they don't have to be redefined every time. But now that I have seen my own code for that, I don't want it anymore :(


```
  getTimeFunc(date) => (Ride ride1, Ride ride2, DateTime date) =>
      (date.difference(ride1.startTime) - date.difference(ride2.startTime)).inMinutes;
  int priceFunc(Ride ride1, Ride ride2) => ((ride1.price! - ride2.price!) * 100).toInt();

  int Function(Ride, Ride) sortFunction(DateTime date) {
    switch (this) {
      case SearchSuggestionSorting.relevance:
        return (ride1, ride2) => getTimeFunc(date)(ride1, ride2) + priceFunc(ride1, ride2);
      case SearchSuggestionSorting.time:
        return getTimeFunc(date);
      case SearchSuggestionSorting.price:
        return priceFunc;
    }
  }
```

lib/rides/pages/search_suggestion_page.dart
Eddie-42: I think we should discuss this in a meeting and set it aside for another user story. I believe we discussed with AG that we should include the time in the search and the algorithm might play a big part in that (is there even an option for the whole day in the algorithm? Or will we have to fire multiple requests to get the results if it only outputs rides with a certain time deviation?) I would personally prefer giving users the option to choose the time (since some trips might be time-critical) but maybe making whole day the default

lib/rides/widgets/search_suggestion_filter.dart
Eddie-42: So you're taking back your own proposal? I'm confused :D I think since you need the date for the timeFunc, it has to be redefined every time regardless, and it would be kind of weird to have those at different places or have one built by a function with an argument, as you proposed. I think performance isn't that much of an issue that we have to save this tiny function declaration (and if the Dart compiler is smart, it saves it anyway)

lib/rides/pages/search_suggestion_page.dart
Eddie-42: Okay I think I understand your request now :D I implemented it and I am not really content with the way all features disappear if you select one in the retracted state, but I couldn't think of a way to fix that while staying true to your proposal...

lib/rides/pages/search_suggestion_page.dart
Eddie-42: Done, now there is just a little bit of ambiguity between clicking the row and dragging through it, but I don't really think that'll be a problem

lib/rides/widgets/search_suggestion_filter.dart
fynsta: Yes but I did not want to have my work got to waste so I just posted how you _would_ do it :D

lib/rides/pages/search_suggestion_page.dart
fynsta: Yeah that is a good point, that does feel pretty weird. Probably would be best to only hide them when going back to the retracted state.

lib/rides/widgets/search_suggestion_filter.dart
```diff
@@ -0,0 +1,449 @@
...
+    return rideSuggestions
+        .where(
+          (Ride ride) {
+            Profile driver = ride.drive!.driver!;
+            bool ratingSatisfied = driver.getAggregateReview().rating >= _minRating &&
+                driver.getAggregateReview().comfortRating >= _minComfortRating &&
+                driver.getAggregateReview().safetyRating >= _minSafetyRating &&
+                driver.getAggregateReview().reliabilityRating >= _minReliabilityRating &&
+                driver.getAggregateReview().hospitalityRating >= _minHospitalityRating;
+            bool featuresSatisfied = Set.of(driver.profileFeatures!).containsAll(_selectedFeatures);
```
istzen, line 408: i don't understand why but if i select any feature there are no more rides shown.

lib/rides/widgets/search_suggestion_filter.dart
```diff
@@ -0,0 +1,449 @@
...
+      widgets.add(featuresRow);
+    }
+    if (!isDeviationDefault) {
+      Widget deviationWidget = Row(
+        children: [const Icon(Icons.schedule), const SizedBox(width: 6), Text("± ${_maxDeviationController.text}")],
+      );
+      widgets.add(deviationWidget);
+    }
+    Widget sortingWidget = Row(
+      children: [const Icon(Icons.sort), Text(_sorting.getDescription(context))],
```
istzen, line 362: can we move this part always to the right side

lib/rides/pages/search_suggestion_page.dart
Eddie-42: Okay, now there is a bit of an **ALGORITHM** that determines which features are shown in retracted state:
- `_selectedFeatures` are always shown, and are always at the top. There is a new variable `_retractedAdditionalFeatures`, the content of which is shown after `_selectedFeatures` (for duplicates, only the `_selectedFeatures` value is shown).
- In the initialization, `_retractedAdditionalFeatures` is set to `_commonFeatures`.
- If you change the state from expanded to retracted and vice versa, `_retractedAdditionalFeatures` is set to `_commonFeatures` if `_selectedFeatures` is empty, otherwise to `[]`. This is pretty much the behavior from before, but having this in the change of state means that features won't disappear instantly.
- If you remove a feature in retracted state, it will be added to the beginning of `_retractedAdditionalFeatures` so that it, too, doesn't disappear instantly, even if it is not in `_commonFeature`.

I think it works quite smoothly.

lib/rides/widgets/search_suggestion_filter.dart
Eddie-42: You mean the entire row other than the filter icon should be aligned to the right or only the sort section or am I misunderstanding you entirely?

lib/rides/widgets/search_suggestion_filter.dart
```diff
@@ -405,7 +416,9 @@ class SearchSuggestionFilter {
                 driver.getAggregateReview().safetyRating >= _minSafetyRating &&
                 driver.getAggregateReview().reliabilityRating >= _minReliabilityRating &&
                 driver.getAggregateReview().hospitalityRating >= _minHospitalityRating;
-            bool featuresSatisfied = Set.of(driver.profileFeatures!).containsAll(_selectedFeatures);
+            List<Feature> driverFeatures =
+                driver.profileFeatures!.map((profileFeature) => profileFeature.feature).toList();
```
fynsta, line 420: let's make this a getter on profile

lib/rides/widgets/search_suggestion_filter.dart
fynsta: Pretty sure he means just the sort section. I agree that it would look better!

lib/rides/pages/search_suggestion_page.dart
fynsta: I like it a lot! The only thing I was wondering: Do we want to keep the order when expanding? That would stop the features from jumping around even more.

Eddie-42: What I can do is sort `_retractedAdditionalFeatures` so that they are in the same order as in the expanded list. Otherwise, I think what you mean is having the `_retractedAdditionalFeatures` in front of the others in the expanded view as well, and I decided against that because I think it would be really confusing if you could just completely destroy the order like that

lib/rides/widgets/search_suggestion_filter.dart
Eddie-42: So it would not be scrollable anymore and glued to the right like the filter icon is glued to the left? I think that would be a bad decision because then the filter icon would become confusing because they look like they should be doing the same, but one indicates what the entire row is for and the other indicates a filter.

lib/rides/pages/search_suggestion_page.dart
fynsta: Yeah I thought that too, I guess that's a matter of preference. Keeping the order but not in front seems unnecessary to me? Let's just keep it like it is.

lib/rides/widgets/search_suggestion_filter.dart
fynsta: Okay but it's the only thing that is always shown. And right now, if there are no other filters, it kind of floats weirdly in the middle (but not symmetrically) of the screen

Eddie-42: I can align the whole thing to the right?

Eddie-42: Did that

istzen (approved):

fynsta (approved):

lib/l10n/app_en.arb
```diff
@@ -302,7 +305,22 @@
...
+  "searchSuggestionsFilterFeaturesSelectTooltip": "Select",
+  "searchSuggestionsFilterFeatures": "Features",
+  "searchSuggestionsFilterFeaturesDeselectTooltip": "Deselect",
+  "searchSuggestionsFilterDeviation": "Maximum deviation from selected time",
+  "searchSuggestionsFilterMinimumRating": "Minimum rating",
+  "searchSuggestionsFilterDeviationHours": "hours",
+  "searchSuggestionsFilterSorting": "Sorting",
+  "searchSuggestionsFilterResetToDefault": "Reset to default",
+  "searchSuggestionsSortingRelevance": "Relevance",
+  "searchSuggestionsSortingTime": "Time",
```
Showcas, line 322: I would call "Time" "Traveltime". To me its not intuitive what Time mean (Starttime, Endtime, Traveltime)

Showcas (approved):

lib/rides/widgets/search_suggestion_filter.dart
```diff
@@ -419,14 +419,11 @@ class SearchSuggestionFilter {
...
             AggregateReview driverReview = driver.getAggregateReview();
-            bool ratingSatisfied = driver.getAggregateReview().rating >= _minRating &&
-                (driverReview.comfortRating >= _minComfortRating ||
-                    _minComfortRating == 1 && !driverReview.isComfortSet) &&
-                (driverReview.safetyRating >= _minSafetyRating || _minSafetyRating == 1 && !driverReview.isSafetySet) &&
-                (driverReview.reliabilityRating >= _minReliabilityRating ||
-                    _minReliabilityRating == 1 && !driverReview.isReliabilitySet) &&
-                (driverReview.hospitalityRating >= _minHospitalityRating ||
-                    _minHospitalityRating == 1 && !driverReview.isHospitalitySet);
+            bool ratingSatisfied = (driverReview.rating >= _minRating || !driverReview.isRatingSet) &&
```
fynsta, line 422: I would switch these conditions. This current way the code relies on the internal behaviour that ("rating set" == "rating == 0"). But that part might change in the future.

PR #51, opened 2023-01-14
Cards

Showcas:
Cards:
- Design changes
- Refresh on changes of the trip
- Notifikation icons
- Greyout when trips are cancelled
- You can soft delete trips when they are Cancelled/rejected (in the Detailed page)

Search Suggestion does not show old drives and is start and end location depended (not done but better dummy, for better testing)

Pending Cards design fix

coauther @istzen


lib/drives/models/drive.dart
```diff
@@ -151,4 +151,19 @@ class Drive extends Trip {
   String toString() {
     return 'Drive{id: $id, from: $start at $startTime, to: $end at $endTime, by: $driverId}';
   }
+
+  @override
+  bool isDifferentFrom(Trip other) {
```
fynsta, line 156: this should be`equals` with turned around semantics

lib/drives/models/drive.dart
```diff
@@ -14,39 +14,39 @@ class Drive extends Trip {
...
+      required super.startTime,
+      required super.end,
+      required super.endPosition,
+      required super.endTime,
+      required super.seats,
+      this.cancelled = false,
+      required this.driverId,
+      this.driver,
+      this.rides,
+      super.hideinListView});
```
fynsta, line 31: please add the trailing comma here and everywhere else, this creates a lot of unnecessary whitespace changes and a worse format

lib/l10n/app_de.arb
```diff
@@ -250,6 +257,8 @@
   "pageRideDetailBannerRejected": "Diese Mitfahrt wurde abgelehnt.",
   "pageRideDetailBannerCancelledByDriver": "Diese Mitfahrt wurde abgesagt.",
   "pageRideDetailBannerCancelledByYou": "Du hast diese Mitfahrt abgesagt.",
+  "pageRideDetailDeleteRide": "Mitfahrt löschen",
+  "pageRideDetailDeteRideDialog": "Wollen Sie wirklich die Mitfahrt löschen?",
```
istzen, line 261: Hier bitte duzen wie bei den anderen Texten

.vscode/settings.json
```diff
@@ -17,5 +17,6 @@
   "files.watcherExclude": {
     "**/.fvm": true
   },
-  "java.configuration.updateBuildConfiguration": "automatic"
+  "java.configuration.updateBuildConfiguration": "automatic",
+  "cmake.sourceDirectory": "${workspaceFolder}/linux"
```
fynsta, line 21: this change should not be in here, right?

lib/rides/models/ride.dart
```diff
@@ -127,6 +127,23 @@ class Ride extends Trip {
     return Ride.fromJsonList(await supabaseClient.from('rides').select().eq('rider_id', userId));
   }
 
+  @override
+  bool isDifferentFrom(Trip other) {
+    if (other is! Ride) return true;
+    Ride ride = other;
+    return start != ride.start ||
```
istzen, line 134: you can also call super.isDifferentFrom (equals) for the parameters that are in trip

lib/drives/pages/drive_detail_page.dart
```diff
@@ -318,6 +320,39 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
     return list;
   }
 
+  void noShowDrive() async {
```
fynsta, line 323: hideDrive?

lib/util/trip/trip.dart
```diff
@@ -9,6 +9,7 @@ abstract class Trip extends Model {
   final String end;
   final Position endPosition;
   final DateTime endTime;
+  final bool hideinListView;
```
istzen, line 12: the i in "in" should be a Uppercase letter

lib/rides/models/ride.dart
```diff
@@ -65,22 +65,22 @@ class Ride extends Trip {
...
+        endPosition: Position(json['end_lat'].toDouble(), json['end_lng'].toDouble()),
+        endTime: DateTime.parse(json['end_time']),
+        seats: json['seats'],
+        price: json['price'],
+        status: RideStatus.values[json['status']],
+        riderId: json['rider_id'],
+        rider: json.containsKey('rider') ? Profile.fromJson(json['rider']) : null,
+        driveId: json['drive_id'],
+        drive: json.containsKey('drive') ? Drive.fromJson(json['drive']) : null,
+        hideinListView: json['hide_in_listview']);
```
istzen, line 83: hide_in_listview -> hide_in_list_view

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -108,10 +108,16 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
 
   //todo: get possible Rides from Algorithm
   void loadRides() async {
+    int riderid = SupabaseManager.getCurrentProfile()?.id ?? -1;
```
fynsta, line 111: the I should be uppercase, because it starts a new word.

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -108,10 +108,16 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
     List<dynamic> data = await supabaseClient
         .from('drives')
         .select('*, driver:driver_id (*)')
         .eq('start', _startController.text)
+        .eq('end', _destinationController.text)
```
fynsta, line 116: let's remove all these filters, because they are present in Simon's PR?

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -277,18 +283,23 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
...
-              separatorBuilder: (BuildContext context, int index) {
-                return const SizedBox(height: 10);
-              },
-              itemCount: _rideSuggestions!.length,
-            ),
-          );
+        : _rideSuggestions!.isEmpty
+            ? Padding(
+                padding: const EdgeInsets.all(8.0),
+                child: Center(child: Text(S.of(context).pageSearchSuggestionsNoRidesFound)),
```
fynsta, line 289: I would also remove this because it's done in Simons PR

lib/rides/models/ride.dart
fynsta: I changed that so we can unblock the admin PR (btw quick plug, please review!!)

lib/rides/pages/search_suggestion_page.dart
Eddie-42: I mean this is basically an algorithm change and not in my PR, but I don't see why it's necessary because it completely removes the possibility of partial rides

Eddie-42 (changes requested):
Got this error when loading my past rides, past drives or searching rides:
```
E/flutter ( 5596): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: setState() called after dispose(): _DriveCard#c917a(lifecycle state: defunct, not mounted)
E/flutter ( 5596): This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent 
widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.
E/flutter ( 5596): The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
E/flutter ( 5596): This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().
E/flutter ( 5596): #0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1078:9)
E/flutter ( 5596): #1      State.setState (package:flutter/src/widgets/framework.dart:1113:6)
E/flutter ( 5596): #2      _DriveCard.loadDrive (package:motis_mitfahr_app/util/trip/drive_card.dart:51:5)
E/flutter ( 5596): <asynchronous suspension>
```

Also, a ProgressIndicator stays forever in such a list and the card just doesn't load:

image

lib/util/trip/drive_card.dart
```diff
@@ -1,76 +1,118 @@
...
+    super.initState();
+
+    setState(() {
+      drive = widget.trip;
+      super.trip = drive;
+    });
+    loadDrive();
+  }
+
+  // todo: changes when rides.status changes
```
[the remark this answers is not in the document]
istzen: i think so. maybe the todo should be removed

lib/util/trip/ride_card.dart
```diff
@@ -1,85 +1,150 @@
...
   const RideCard(super.trip, {super.key});
 
+  @override
+  State<RideCard> createState() => RideCardState();
+}
+
+class RideCardState extends TripCardState<RideCard> {
+  Ride? ride;
+  bool fullyLoaded = false;
+  Profile? driver;
```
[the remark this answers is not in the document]
Showcas: Agreed. Final is sadly not possible since when there are changes it reloads and they get reassigned

lib/util/trip/trip_card_state.dart
```diff
@@ -0,0 +1,116 @@
...
+          children: [
+            buildBottomLeft(),
+            buildBottomRight(),
+          ],
+        ),
+      ],
+    );
+  }
+
+  Widget buildBottomLeft();
```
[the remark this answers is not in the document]
Showcas: sind die das nicht?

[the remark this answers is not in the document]
fynsta: Ja stimmt das war mein Fehler!

lib/util/trip/drive_card.dart
```diff
@@ -1,76 +1,118 @@
...
-                      ),
-                    ),
-                    node: const TimelineNode(
-                      indicator: CustomOutlinedDotIndicator(),
-                      startConnector: CustomSolidLineConnector(),
-                    ),
-                  )
-                ],
+    return !fullyLoaded
+        ? const Center(child: CircularProgressIndicator())
```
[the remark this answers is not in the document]
Showcas: > Got this error when loading my past rides, past drives or searching rides:
the error is fixed.

> Also, a ProgressIndicator stays forever in such a list and the card just doesn't load:
I couldn't replicate the ProgressIndicator issue. Could you test if it is fixed?
(Note: I removed the progressionIndicators so When you load the page you don't have all of them a you requested)
For looking if your problem still exists, change the Sizedbox after fullybuild check in the build method of the Card to the Indicator for testing

Showcas:
> Got this error when loading my past rides, past drives or searching rides:
> 
> ```
> E/flutter ( 5596): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: setState() called after dispose(): _DriveCard#c917a(lifecycle state: defunct, not mounted)
> E/flutter ( 5596): This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent 
> widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.
> E/flutter ( 5596): The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
> E/flutter ( 5596): This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().
> E/flutter ( 5596): #0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1078:9)
> E/flutter ( 5596): #1      State.setState (package:flutter/src/widgets/framework.dart:1113:6)
> E/flutter ( 5596): #2      _DriveCard.loadDrive (package:motis_mitfahr_app/util/trip/drive_card.dart:51:5)
> E/flutter ( 5596): <asynchronous suspension>
> ```
> 
> Also, a ProgressIndicator stays forever in such a list and the card just doesn't load:
> 
> image

the error is fixed.

I couldn't replicate the ProgressIndicator issue. Could you test if it is fixed?
(Note: I removed the progressionIndicators so When you load the page you don't have all of them a you requested)
For looking if your problem still exists, change the Sizedbox after fullybuild check in the build method of the Card to the Indicator for testing

Eddie-42:
Okay this behavior seems to be fixed

lib/drives/pages/drive_chat_page.dart
```diff
@@ -79,6 +79,7 @@ class _DriveChatPageState extends State<DriveChatPage> {
       body: RefreshIndicator(
         onRefresh: loadDrive,
         child: ListView.separated(
+          padding: const EdgeInsets.all(8),
```
istzen, line 82: Why is this in here?

Eddie-42:
Would it be possible to implement the following behavior: More cards (or all of them) are loaded initially and until they are all loaded, there is a singular ProgressIndicator in the center. After that, cards are loaded earlier when I go down the list and maybe cached when I go up. I am really annoyed at how badly I can scroll through the list right now, since every two cards or so you can see how the page loads the next cards for about 1s, which adds up. I know that this is a stream thing, but I'm sure there is a better way

lib/util/trip/drive_card.dart
```diff
@@ -1,76 +1,109 @@
...
-          ),
-          child: Column(
-            crossAxisAlignment: CrossAxisAlignment.start,
-            children: [
-              Padding(
-                padding: const EdgeInsets.all(12.0),
-                child: Text(localeManager.formatDate(trip.startTime)),
+    return !fullyLoaded
+        ? const Center(child: SizedBox())
+        : Card(
```
istzen, line 97: Die ausgegrauten Cards sind jz bisschen komisch, weil die zumindest im Light-Mode fast mehr hervorgehoben sind als die anderen. kannst sowas wie das versuchen, dann ist wirklich die komplette Card ausgegraut:
```
Container(
  foregroundDecoration: BoxDecoration(
    color: Theme.of(context).disabledColor.withOpacity(0.05),
    backgroundBlendMode: BlendMode.saturation,
  ),
  child: child,
)
```

.vscode/settings.json
```diff
@@ -17,5 +17,4 @@
   "files.watcherExclude": {
     "**/.fvm": true
   },
-  "java.configuration.updateBuildConfiguration": "automatic"
```
istzen, line 20: Why is this here?

Showcas: It shouldn't be there thank you

Showcas:
> Would it be possible to implement the following behavior: More cards (or all of them) are loaded initially and until they are all loaded, there is a singular ProgressIndicator in the center. After that, cards are loaded earlier when I go down the list and maybe cached when I go up. I am really annoyed at how badly I can scroll through the list right now, since every two cards or so you can see how the page loads the next cards for about 1s, which adds up. I know that this is a stream thing, but I'm sure there is a better way

I will look into it again. But to my knowledge it is at the moment we use Streams or the Cards are all preloaded. But what I could maybe do is catch the Stream before and create the Cards with a List and then trigger the update in a higher tier. The only issue with that is that all trips will readload and not only the one that changed

lib/util/trip/drive_card.dart
Showcas: Das hatte ich da vor auch verwendet aber dadurch sind diese grauen ecken entstanden.  Wenn es eine Möglichkeit gibt die ecken auf die Card zu matchen dann gerne aber ich habe keine Methode gefunden.  Außerdem würde bei dem das rote Block Icon auch saturiert werden, weshalb ich es damals dann auf einen Stack gepackt habe.

Eddie-42:
What I think you could do quite easily is not wait until the entire card is loaded to display it, and load most (or all?) of the information directly in the stream supabase query

Eddie-42:
Was haltet ihr von diesem Design, um den Status der Fahrt mit dem Streifen anzuzeigen? (Rot=Abgelehnt/Abgesagt, Blau=Angefragt, Grün=Angenommen?): https://stackoverflow.com/questions/68209584/flutter-card-and-container-design

lib/util/trip/drive_card.dart
Eddie-42: Hattest du die Card als child vom Container? Weil ich mal versuchen würde, den Container als child der Card zu machen und den ganzen Inhalt als child des Container

Showcas:
> What I think you could do quite easily is not wait until the entire card is loaded to display it, and load most (or all?) of the information directly in the stream supabase query

I moved the fullyloaded blocker back, so the Widgets that need to wait are loaded later so the Cards can already be seen. Do you like this or do you prefer it different? I think this is the fastest way for the Cards to appear.

Showcas:
> Was haltet ihr von diesem Design, um den Status der Fahrt mit dem Streifen anzuzeigen? (Rot=Abgelehnt/Abgesagt, Blau=Angefragt, Grün=Angenommen?): https://stackoverflow.com/questions/68209584/flutter-card-and-container-design

wäre auch nice. vielleicht schaff ich dort dann auch das mit der Saturation richtig zu kombinieren. Ein versuch ist es wert

fynsta:
> Was haltet ihr von diesem Design, um den Status der Fahrt mit dem Streifen anzuzeigen? (Rot=Abgelehnt/Abgesagt, Blau=Angefragt, Grün=Angenommen?): https://stackoverflow.com/questions/68209584/flutter-card-and-container-design

Finde ich geil, so hatte ich mir das auch vorgestellt (oder halt Streifen über die Ecke aber da sind ja jetzt Infos)

Showcas:
Hab jetzt diese Banner eingebaut Drive: (Grün = takes place / Red = Cancelled)) Ride: (Grün = acceppted / Red = Cancelled,rejected / Grey = pending). Bei light mode ist die graue Karte gut zu sehen aber in Darkmode kommt der Greyeffektnicht wirklich zur Geltung.

lib/util/trip/drive_card.dart
```diff
@@ -70,43 +70,69 @@ class _DriveCardState extends TripCardState<DriveCard> {
...
-            : drive!.rides!.any((ride) => ride.status == RideStatus.pending)
-                ? Icon(
-                    Icons.done_all,
-                    color: Theme.of(context).disabledColor,
-                  )
-                : Icon(
-                    Icons.done_all,
-                    color: Theme.of(context).own().success,
-                  );
+    return const SizedBox();
```
fynsta, line 84: maybe the default in the TripCardState should just be SizedBox(). So you don't need to implement it when the card is not doing anything with that space

lib/util/trip/drive_card.dart
```diff
@@ -70,43 +70,69 @@ class _DriveCardState extends TripCardState<DriveCard> {
...
 
+  Color pickBannerColor() {
+    if (_drive!.cancelled) {
+      return Theme.of(context).errorColor;
+    } else {
+      return Theme.of(context).own().success;
+    }
+  }
+
   // Notification
```
fynsta, line 81: Comment is not accurate anymore

lib/util/trip/drive_card.dart
```diff
@@ -70,43 +70,69 @@ class _DriveCardState extends TripCardState<DriveCard> {
...
-                : Icon(
-                    Icons.done_all,
-                    color: Theme.of(context).own().success,
-                  );
+    return const SizedBox();
   }
 
-  Color pickColor() {
-    return drive!.cancelled ? Theme.of(context).disabledColor.withOpacity(0.05) : Theme.of(context).cardColor;
+  BoxDecoration pickDecoration() {
```
fynsta, line 87: the borderRadius is the same for both, so I would move that to a variable

lib/util/trip/ride_card.dart
```diff
@@ -145,23 +133,75 @@ class _RideCardState extends TripCardState<RideCard> {
...
+  Color pickBannerColor() {
+    switch (_ride!.status) {
+      case RideStatus.preview:
+        return Theme.of(context).cardColor;
+      case RideStatus.pending:
+        return Theme.of(context).disabledColor;
+      case RideStatus.approved:
+        return Theme.of(context).own().success;
+      case RideStatus.rejected:
+        return Theme.of(context).errorColor;
```
fynsta, line 145: Somehow `errorColor` is different than `colorScheme.error`. No idea why, but we always used the latter because it is more accurate. So please change that for your PR

lib/util/trip/drive_card.dart
Showcas: Found a sollution

.vscode/settings.json
fynsta: No? The line was there because it is useful, and you removed it from the file

lib/drives/pages/drive_detail_page.dart
```diff
@@ -181,16 +181,18 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
         widgets.addAll(pendingRidesColumn);
       }
 
+      widgets.add(const SizedBox(height: 10));
+      Widget bottomButton;
       if (!(_drive!.isFinished || _drive!.cancelled)) {
```
fynsta, line 186: Can you switch around the if/else cases? would be easier to read

lib/l10n/app_de.arb
```diff
@@ -247,6 +247,9 @@
   "pageDriveDetailCancelDialogMessage": "Bist du sicher, dass du diese Fahrt absagen möchtest?",
   "pageDriveDetailCancelDialogToast": "Fahrt abgesagt.",
   "pageDriveDetailBannerCancelled": "Du hast diese Fahrt abgesagt.",
+  "pageDriveDetailDeleteDrive": "Fahrt löschen",
+  "pageDriveDetailDeleteDriveDialog": "Wollen Sie wirklich die Fahrt löschen?",
```
fynsta, line 251: come on - please stay with our style:
1. Duzen!!!!!
2. "Bist du sicher, dass ..."

lib/l10n/app_en.arb
```diff
@@ -386,6 +389,7 @@
 
   "pageSearchSuggestionsTitle": "Search Suggestions",
   "pageSearchSuggestionsButtonFilter": "Filter",
+  "pageSearchSuggestionsNoRidesFound": "There are no fitting rides available",
```
fynsta, line 392: unused translation

lib/rides/pages/ride_detail_page.dart
```diff
@@ -189,7 +189,10 @@ class _RideDetailPageState extends State<RideDetailPage> {
           onPressed: _showWithdrawDialog,
         );
       default:
```
fynsta, line 191: would probably be better to make this case explicit (which statuses apply?)

lib/util/trip/drive_card.dart
```diff
@@ -1,75 +1,121 @@
...
+      *,
+      rides(
+        *,
+        rider: rider_id(*)
+      )
+    ''').eq('id', widget.trip.id).single();
+    if (mounted) {
+      setState(() {
+        _drive = Drive.fromJson(data);
+        super.trip = _drive;
```
fynsta, line 52: `super` here and everywhere else is unneeded, it's implied

lib/util/trip/drive_card.dart
```diff
@@ -1,75 +1,121 @@
...
-                  TimelineTile(
-                    contents: Padding(
-                      padding: const EdgeInsets.all(32.0),
-                      child: Row(
-                        mainAxisAlignment: MainAxisAlignment.spaceBetween,
-                        children: [
-                          Text('${localeManager.formatTime(trip.endTime)}  ${trip.end}'),
-                        ],
+            margin: const EdgeInsets.only(left: 10),
+            child: InkWell(
```
fynsta, line 108: this InkWell does not work anymore

fynsta: probably should be around the Wrap element?

lib/util/trip/drive_card.dart
```diff
@@ -1,75 +1,121 @@
...
+
+  @override
+  Widget buildRightSide() {
+    return TripOverview(super.trip!).buildSeatIndicator(context, super.trip!);
+  }
+
+  Color pickBannerColor() {
+    if (!_fullyLoaded) {
+      return Theme.of(context).cardColor;
+    } else {
```
fynsta, line 66: past drives should not follow this color scheme? they should be always green, for example

lib/util/trip/drive_card.dart
```diff
@@ -1,75 +1,121 @@
...
+      if (_drive!.cancelled) {
+        return Theme.of(context).colorScheme.error;
+      } else if (_drive!.rides!.any((ride) => ride.status == RideStatus.pending)) {
```
fynsta, line 69: why would it be disabled in this case? It could maybe be yellow or blue or something (primary for example)

lib/util/trip/drive_card.dart
```diff
@@ -1,75 +1,121 @@
...
+        return Theme.of(context).disabledColor;
+      } else {
+        return Theme.of(context).own().success;
```
fynsta, line 72: maybe this should be disabledColor if there aren't any rides or ride requests?

lib/util/trip/ride_card.dart
```diff
@@ -1,84 +1,195 @@
...
+      );
+    }
+  }
+
+  Color pickBannerColor() {
+    switch (_ride!.status) {
+      case RideStatus.preview:
+        return Theme.of(context).cardColor;
+      case RideStatus.pending:
+        return Theme.of(context).disabledColor;
```
fynsta, line 141: this should be warning to comply with the banner

lib/util/trip/ride_card.dart
```diff
@@ -1,84 +1,195 @@
...
+      shape: RoundedRectangleBorder(
+        borderRadius: BorderRadius.circular(10),
+      ),
+      child: Wrap(
+        children: [
+          Container(
+            foregroundDecoration: pickDecoration(),
+            width: MediaQuery.of(context).size.width,
+            decoration: BoxDecoration(
+              color: Theme.of(context).cardColor,
```
fynsta, line 180: this does not seem to work, the whole card is white

lib/util/trip/trip_card_state.dart
```diff
@@ -0,0 +1,131 @@
...
+            child: Row(
+              mainAxisAlignment: MainAxisAlignment.spaceBetween,
+              children: [
+                SizedBox(
+                  child: Row(
+                    children: [
+                      const Icon(Icons.access_time_outlined),
+                      const SizedBox(width: 4),
+                      Text(
+                          "${trip!.duration.inHours.toString().padLeft(2, "0")}:${(trip!.duration.inMinutes % 60).toString().padLeft(2, "0")}"),
```
fynsta, line 51: can you move this to locale manager? `formatDuration`
And it should probably have a parameter to control whether the hours are padded. Here for the duration it makes sense, but I don't think it makes sense for the added duration of a ride

fynsta:
I have an overflow,  presumably because of the long name
image

lib/util/trip/ride_card.dart
```diff
@@ -1,84 +1,195 @@
...
+      case RideStatus.cancelledByRider:
+        return Theme.of(context).colorScheme.error;
+      case RideStatus.withdrawnByRider:
+        return Theme.of(context).cardColor;
+    }
+  }
+
+  BoxDecoration pickDecoration() {
+    if (_ride!.status.isCancelled() || _ride!.status == RideStatus.rejected) {
+      return BoxDecoration(
```
fynsta, line 157: this looks okay-ish in light mode, but have tried out dark mode? No thank you. I think we need to find a better way here to grey out the card

lib/rides/pages/ride_detail_page.dart
Showcas: .cancelledbyDriver, .cancelledbyRider and .rejected

fynsta: yeah so I would put those cases explicitly in the code as case statements

lib/util/trip/ride_card.dart
Showcas: Do you mean when the card is disabled or the normal cards?

fynsta: the normal cards, see here:

image

lib/util/trip/drive_card.dart
Showcas: It InkWells the last ancestor normally. If I put before the Wrap it inkwells the Bannerbackgroundcard. Im still trying to figure out how to get it into the front layer.

fynsta: >  If I put before the Wrap it inkwells the Bannerbackgroundcard
How does that look? Can you send a picture?

lib/util/trip/ride_card.dart
fynsta: Card color is fine, but it should be applied to the whole card.

lib/util/trip/drive_card.dart
Showcas: Screenshot 2023-01-18 145539

You don't really see it but the grey is different in this example

Showcas: I think if we want an Inkwell that is the best I can do at the momement :/

fynsta: Simon and I can try to fix that, maybe we will find something

Eddie-42: Done.

.vscode/settings.json
fynsta: Can you please add the line back?

lib/drives/pages/drive_detail_page.dart
```diff
@@ -181,16 +181,21 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
       }
 
-      if (!(_drive!.isFinished || _drive!.cancelled)) {
-        widgets.add(const SizedBox(height: 10));
-        Widget deleteButton = Button(
+      widgets.add(const SizedBox(height: 10));
+      Widget bottomButton;
+      if (_drive!.isFinished || _drive!.cancelled) {
+        bottomButton = Button.error(
+          S.of(context).pageDriveDetailHideDrive,
```
fynsta, line 188: If we want to be consistent, the key for this should be `pageDriveDetailButtonHide`

lib/l10n/app_de.arb
```diff
@@ -304,6 +307,8 @@
   "pageRideDetailBannerRejected": "Diese Mitfahrt wurde abgelehnt.",
   "pageRideDetailBannerCancelledByDriver": "Diese Mitfahrt wurde abgesagt.",
   "pageRideDetailBannerCancelledByYou": "Du hast diese Mitfahrt abgesagt.",
+  "pageRideDetailHideRide": "Mitfahrt löschen",
+  "pageRideDetailHideRideDialog": "Willst du wirklich die Mitfahrt löschen?",
```
fynsta, line 311: Auch hier "Bist du sicher"

lib/util/trip/drive_card.dart
```diff
@@ -1,74 +1,141 @@
...
+  @override
+  Widget buildRightSide() {
+    return TripOverview(super.trip!).buildSeatIndicator(context, trip!);
+  }
+
+  Color pickBannerColor() {
+    if (!_fullyLoaded) {
+      return Theme.of(context).cardColor;
+    } else if (_drive!.endTime.isBefore(DateTime.now())) {
+      if (_drive!.cancelled) {
```
fynsta, line 69: Hm I would either
1. always show disabled
2. or show success only for `approved` rides

MarkFalke (approved):

fynsta:
We refactored the cards so that all of them use the `TripCard` default layout, and so that there is as little duplicated code as possible. 
The pending rides still look a little bit weird, we will do another pass over them when we add rider reviews because that will change the design again anyways.

fynsta (approved):

Eddie-42 (approved):

PR #53, opened 2023-01-14
refactor pending rides

istzen:
as we discussed in our last Meeting i moved the pending rides from the driveChatPage to the driveDetailView.

### DIFF
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -227,11 +227,11 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
             ).then((value) => loadDrive()),
             icon: Badge(
               badgeContent: Text(
-                _getMessageCount(_drive!) == 0 ? '' : _getMessageCount(_drive!).toString(),
+                _getMessageCount(_drive!).toString(),
                 style: const TextStyle(color: Colors.white),
                 textScaleFactor: 1.0,
               ),
-              showBadge: _getMessageCount(_drive!) != null,
+              showBadge: _getMessageCount(_drive!) != 0,
               position: BadgePosition.topEnd(top: -12),
               child: const Icon(Icons.chat),
             ),
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -176,9 +176,8 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
             ),
           ),
           const SizedBox(height: 10.0),
+          ..._pendingRidesList(pendingRides)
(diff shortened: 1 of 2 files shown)


lib/drives/pages/drive_detail_page.dart
```diff
@@ -353,6 +371,25 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
     await _drive?.cancel();
     setState(() {});
   }
+
+  int? _getMessageCount(Drive drive) {
```
fynsta, line 375: how would that be null? that makes no sense to me... If there are no messages, then it should be zero.

lib/drives/pages/drive_detail_page.dart
```diff
@@ -163,6 +164,23 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
+          Align(
+            alignment: Alignment.centerLeft,
+            child: Text(
+              S.of(context).pageDriveChatRequestsHeadline,
+              style: Theme.of(context).textTheme.titleLarge,
+            ),
+          ),
+          const SizedBox(height: 10.0),
+        ];
+        pendingRidesColumn.addAll(_pendingRidesList(pendingRides));
```
fynsta, line 180: you don't need `addAll` here, you could put  `... _pendingRidesList(pendingRides)` in the last line of the column array definition

lib/drives/pages/drive_detail_page.dart
istzen: yeah this method right now is just a placeholder the real method will be provided when ich finish the chat. I'll leave it for now if that's okay with you

lib/drives/pages/drive_detail_page.dart
istzen: thanks for the hint, and sorry that this took two commits.

lib/drives/pages/drive_detail_page.dart
fynsta: sure but the placeholder can still be accurate -> i.e. it cannot return null so the signature is wrong

istzen: yes that right i changed it

fynsta (approved):

Eddie-42 (approved):

MarkFalke (approved):

fynsta:
@istzen So did you want to remove the pendingRides stuff that's still here on this branch? In the chat page

Showcas (approved):

istzen:
> @istzen So did you want to remove the pendingRides stuff that's still here on this branch? In the chat page

yeah i just did it in the chat branch

PR #54, opened 2023-01-15
Better tests

fynsta:
This adds test factories. The test factories look pretty intimidating, so please don't hesitate to ask for clarification if you don't understand something (they are not auto-generated, but this is the standard way we thought of - they should be written exactly the same way in the future).

Furthermore, example tests for `Profile` are added to give everyone a starting point on unit tests. One open TODO is localizations (and context as a whole), we have not figured out a way to do that in the scope of the unit tests where no widgets (and thus no building) is present.

Also adds some small cleanups and improvements that we found along the way:

- adds `toJsonForApi` which also adds `id` and `createdAt`, in comparison to the existing `toJson` method
- removes `toJsonList` because it is not used anywhere and I don't know what we would use it for
- removes `getAggregateReview`, it's unused
- adds Review.maxRating constant for reuse

Co-authored by @Eddie-42


test/models/profile_test.dart
```diff
@@ -0,0 +1,185 @@
...
+    test('returns name and surname if both are set', () async {
+      final profile = ProfileFactory().generateFake(
+        name: NullableParameter('name'),
+        surname: NullableParameter('surname'),
+      );
+      expect(profile.fullName, 'surname name');
+    });
+  });
+
+  group('isCurrentUser', () {
```
istzen, line 62: what if profile.id == null

test/models/profile_test.dart
```diff
@@ -0,0 +1,185 @@
...
+        "username": "username",
+        "email": "email",
+        "description": "description",
+        "birth_date": "2021-01-01T00:00:00.000Z",
+        "surname": "surname",
+        "name": "name",
+        "gender": 0,
+        "avatar_url": "avatar_url",
+      };
+      Profile profile = Profile.fromJson(json);
```
istzen, line 94: is there a way to check if the json contains any other information?

test/models/profile_test.dart
```diff
@@ -0,0 +1,185 @@
...
+      expect(profile.avatarUrl, json["avatar_url"]);
+    });
+
+    test('can handle associated models', () async {
+      Map<String, dynamic> json = {
+        "id": 1,
+        "created_at": "2021-01-01T00:00:00.000Z",
+        "username": "username",
+        "email": "email",
+        "reviews_received": ReviewFactory().generateFakeJsonList(length: 3),
```
istzen, line 113: does this work if the reviews_received == null same for the others?

test/models/profile_test.dart
```diff
@@ -0,0 +1,185 @@
...
+        "reports_received": ReportFactory().generateFakeJsonList(length: 3),
+      };
+      Profile profile = Profile.fromJson(json);
+      expect(profile.reviewsReceived!.length, 3);
+      expect(profile.profileFeatures!.length, 3);
+      expect(profile.reportsReceived!.length, 3);
+    });
+  });
+
+  group('Profile.fromJsonList', () {
```
istzen, line 124: maybe add a test with an empty list

lib/account/models/profile.dart
```diff
@@ -90,10 +91,6 @@ class Profile extends Model {
     };
   }
 
-  List<Map<String, dynamic>> toJsonList(List<Profile> profiles) {
```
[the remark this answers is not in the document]
fynsta: Read the description 😉

test/models/profile_test.dart
fynsta: Do you mean if we used NullableParameter to set them?

test/models/profile_test.dart
```diff
@@ -0,0 +1,185 @@
...
+      expect(json["name"], profile.name);
+      expect(json["gender"], profile.gender?.index);
+      expect(json["avatar_url"], profile.avatarUrl);
+    });
+  });
+
+  group('toString', () {
+    test('returns a string representation of the profile', () async {
+      Profile profile = ProfileFactory().generateFake(
+        id: 1,
```
istzen, line 159: what happens if id or created_at is null?

test/models/profile_test.dart
fynsta: but we just defined the json in the line above? Do you mean the `toJson` method?

test/models/profile_test.dart
fynsta: Good point, I'll do that

test/models/profile_test.dart
istzen: yes if thats the way to do that

test/models/profile_test.dart
fynsta: we _could_ check that but it's not really a relevant case for us - also it does not change anything and does not help our coverage %

test/models/profile_test.dart
fynsta: Okay sorry for my confusion. That would be the way if we used the factory here. But we decided to hard-code the json for these specific tests so we can test independently from the correctness of the `toJson` method.

As to your original question: In this case, null is the default. So this is covered in the test case one above, 'parses a profile from json' where those values are not set.

test/models/profile_test.dart
istzen: yes i'm sorry i meant toJson wrong line.

test/models/profile_test.dart
istzen: yeah that's right my mistake

test/models/profile_test.dart
fynsta: I also think this is an edgecase that should not really happen in normal usage. Also, the ModelFactory pattern currently does not support nullable id and created_at yet (we could add that but probably not worth it).

istzen: i guess same reasoning as above that it's not a relevant case for us.

test/models/profile_test.dart
fynsta: Okay, added a check for length.

istzen (approved):

Showcas (approved):

MarkFalke (approved):

PR #55, opened 2023-01-15
Create LICENSE.md

fynsta:
This will be needed when we publish our app to the app store.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 MOTIS Mitfahr-App
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


Eddie-42:
Where did you get this from/Is this exhaustive and legally safe? Can you include a link to this on the About page and remove the current text?

fynsta:
MIT is the standard license OpenSource projects, basically allows all users to do anything they want with the software. You can get the license via Github itself, there is a dialog to help you add the right license.

Where in the about page should I add this? (This is not the terms of service, just a license for software use)

istzen (approved):

Eddie-42 (approved):

MarkFalke (approved):

Showcas (approved):

PR #58, opened 2023-01-16
change version

istzen:
as i sad a short PR to change the version from 3.3.9 to 3.3.10


fynsta:
I would not have approved this - you also need to change the version in the build pipeline (I'll do it quickly on main)

PR #60, opened 2023-01-16
Organize imports

fynsta:
Just a very quick PR to make our imports consistent. There is no definite consensus on whether relative or absolute is better: https://stackoverflow.com/questions/59693195/flutter-imports-relative-path-or-package.

But I think relative is shorter and more concise. What do you all think? 

Imports should be converted automatically on save (in VSCode). Then I used the new `import_sorter` package to sort the imports into 4 groups: Dart imports, Flutter imports, foreign package imports, and local imports. This can be done using `flutter pub run import_sorter:main --no-comments` (everyone should do it before committing). I also added a step to the build pipeline that checks if the imports are okay.

Please review as quickly as possible so we can merge this and have as little conflicts as possible. Then after rebasing your PRs, please also use the same steps to organize the imports of those PRs.


fynsta:
I'll just quickly also try to organize the imports themselves a little bit more

fynsta:
Imports are organized, I will add the info to the PR description.

Eddie-42 (approved):

Showcas (approved):
looks great. Don't forget to run the commands for this PR aswell :)

fynsta:
Removed the import organiser package again, as I don't think it brings any value. The default organisation is in three groups, which is distinct enough.

istzen (approved):

PR #61, opened 2023-01-16
Fix small bug when leaving report unfinished

fynsta:
I had forgotten that is not typed, its dynamic... so need to check for `true` explicitly.

@Eddie-42 while we're here: I had overflowing text on the report page. Could you maybe fix that as part of this PR?


Eddie-42:
Couldn't you just type it? I think that would be much cleaner

fynsta:
How would you do that?

Eddie-42 (approved):

istzen (approved):

PR #62, opened 2023-01-16
fix approve and reject ride

istzen:
now we can really drop the RLS policy for the rides. Please aslo review the functions in supabase (approve_ride, reject_ride, is_update_allowed)

### DIFF
--- a/lib/rides/models/ride.dart
+++ b/lib/rides/models/ride.dart
@@ -74,7 +74,7 @@ class Ride extends Trip {
       endPosition: Position(json['end_lat'].toDouble(), json['end_lng'].toDouble()),
       endTime: DateTime.parse(json['end_time']),
       seats: json['seats'],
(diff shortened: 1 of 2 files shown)


fynsta (approved):

Showcas (approved):

Eddie-42 (approved):

fynsta:
@istzen wouldn't it be easier (and less functions!) if we just had a set_ride_status method with a status parameter? That way we could also use RideStatus.approved.index in flutter instead of placing that random "2" in the supabase function.

istzen:
> @istzen wouldn't it be easier (and less functions!) if we just had a set_ride_status method with a status parameter? That way we could also use RideStatus.approved.index in flutter instead of placing that random "2" in the supabase function.

Yes that would be easier but would also give the users more rights than nessacary.

PR #64, opened 2023-01-18
Enable avoid_dynamic_calls

fynsta:
Making all of this a little more type-safe.

Also adds a `ParseHelper` for the common difficult cases, i.e. double and List<Map<String, dynamic>>.

(I'm gonna try to enable these rules one-by-one, and see how far we can go with that).


lib/util/parse_helper.dart
```diff
@@ -0,0 +1,16 @@
+ParseHelper parseHelper = ParseHelper();
+
+class ParseHelper {
+  parseDouble(dynamic value) {
+    if (value is double) return value;
+    if (value is int) return value.toDouble();
+    return null;
```
Eddie-42, line 7: I think this should throw an Exception, otherwise it will come later down the line and is less debuggable

Eddie-42 (approved):

fynsta:
I also added the always_declare_return_types rule because it is a fitting counterpart.

lib/rides/pages/ride_detail_page.dart
```diff
@@ -237,7 +237,7 @@ class _RideDetailPageState extends State<RideDetailPage> {
     setState(() {});
   }
 
-  _showRequestDialog() {
+  void _showRequestDialog() {
```
istzen, line 240: why is this nessacery?

istzen: nvm read your comment

fynsta: Because of the always_declare_return_types linting rule. It requires you to always declare the return types.

If you don't declare the return types, then dart thinks it has `dynamic` type. That could lead to problems (probably not in this specific case, but generally).

Showcas (approved):

PR #66, opened 2023-01-18
Always specify types

fynsta:
This is a big one. `always_specify_types` make sure that we will always declare types for every variable we encounter. You can look at the PR to see some common cases where this makes a difference:

- specifying `<String,dynamic>` for the json (for example when we talk to supabase)
- specifying `<Widget>` for Widget-Lists, for example in rows and columns
- generally all calls for `List.generate` and `[]` should signify the type of lists
- specifying the return type of `MaterialPageRoute` (mostly void, except if you need the response)
- `Future.delayed` becomes `Future<void>.delayed`

While doing this, I found some problems:
- some `fromJsonList` methods had weird parameters
- we used a dynamic `context` in our Edit...Pages. Now that flutter knows that it is a `BuildContext`, it showed the age-old problem with `context` across async gaps. So I had to change those pages to `Stateful` to get access to `mounted` (but I did see that flutter will have an easier `context.mounted`in the future: https://github.com/flutter/flutter/issues/111488. The annoying thing is that it's already in the docs, but not released yet: https://api.flutter.dev/flutter/widgets/BuildContext/mounted.html)

I disabled the lint rule in our tests, because I think it's more annoying than helpful in those files.

### DIFF
--- a/lib/account/pages/reviews_page.dart
+++ b/lib/account/pages/reviews_page.dart
@@ -3,6 +3,7 @@ import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 
 import '../../rides/models/ride.dart';
 import '../../util/buttons/button.dart';
+import '../../util/parse_helper.dart';
 import '../../util/profiles/profile_widget.dart';
 import '../../util/profiles/reviews/aggregate_review_widget.dart';
 import '../../util/supabase.dart';
@@ -47,10 +48,12 @@ class _ReviewsPageState extends State<ReviewsPage> {
         *,
         writer: writer_id(*)
       )''').eq('id', _profileId).single();
-    List<Map<String, dynamic>> commonRidesData = await SupabaseManager.supabaseClient.from('rides').select('''
+    List<Map<String, dynamic>> commonRidesData = parseHelper.parseListOfMaps(
+      await SupabaseManager.supabaseClient.from('rides').select('''
           *,
           drive: drives!inner(*)
-        ''').eq('rider_id', SupabaseManager.getCurrentProfile()!.id).eq('drive.driver_id', _profileId);
+        ''').eq('rider_id', SupabaseManager.getCurrentProfile()!.id).eq('drive.driver_id', _profileId),
+    );
 
     setState(() {
       _profile = Profile.fromJson(profileData);
--- a/lib/account/pages/reviews_page.dart
+++ b/lib/account/pages/reviews_page.dart
@@ -47,7 +47,7 @@ class _ReviewsPageState extends State<ReviewsPage> {
         *,
         writer: writer_id(*)
       )''').eq('id', _profileId).single();
-    List<dynamic> commonRidesData = await SupabaseManager.supabaseClient.from('rides').select('''
+    List<Map<String, dynamic>> commonRidesData = await SupabaseManager.supabaseClient.from('rides').select('''
           *,
           drive: drives!inner(*)
         ''').eq('rider_id', SupabaseManager.getCurrentProfile()!.id).eq('drive.driver_id', _profileId);
@@ -55,8 +55,8 @@ class _ReviewsPageState extends State<ReviewsPage> {
     setState(() {
       _profile = Profile.fromJson(profileData);
       List<Ride> rides = Ride.fromJsonList(commonRidesData);
-      _hasReviewed = _profile!.reviewsReceived!.any((review) => review.writer!.isCurrentUser);
-      _reviewable = rides.any((ride) => ride.isFinished && ride.status == RideStatus.approved);
+      _hasReviewed = _profile!.reviewsReceived!.any((Review review) => review.writer!.isCurrentUser);
+      _reviewable = rides.any((Ride ride) => ride.isFinished && ride.status == RideStatus.approved);
     });
   }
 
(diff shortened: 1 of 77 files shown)


fynsta:
I forgot to add: Please click through the app and see if I broke anything. I don't think I did, but you never know.

istzen:
the contact us button is not working for me but i'm not sure if that worked before for ios

lib/account/pages/reviews_page.dart
```diff
@@ -47,16 +47,16 @@ class _ReviewsPageState extends State<ReviewsPage> {
         *,
         writer: writer_id(*)
       )''').eq('id', _profileId).single();
-    List<dynamic> commonRidesData = await SupabaseManager.supabaseClient.from('rides').select('''
+    List<Map<String, dynamic>> commonRidesData = await SupabaseManager.supabaseClient.from('rides').select('''
```
istzen, line 50: the following Exception Occours here
>  Unhandled Exception: type 'List<dynamic>' is not a subtype of type 'List<Map<String, dynamic>>'

lib/rides/pages/search_suggestion_page.dart
```diff
@@ -113,17 +113,17 @@ class _SearchSuggestionPage extends State<SearchSuggestionPage> {
 
   //todo: get possible Rides from Algorithm
   Future<void> loadRides() async {
-    List<dynamic> data = await SupabaseManager.supabaseClient.from('drives').select('''
+    List<Map<String, dynamic>> data = await SupabaseManager.supabaseClient.from('drives').select('''
```
istzen, line 116: the Exception is here aswell

Unhandled Exception: type 'List' is not a subtype of type 'List<Map<String, dynamic>>'

lib/account/pages/reviews_page.dart
fynsta: Thanks, fixed both of those!

fynsta:
> the contact us button is not working for me but i'm not sure if that worked before for ios

Yeah I don't think ever worked on ios: https://stackoverflow.com/questions/25981422/how-to-open-mail-app-from-swift#comment47814121_25991205

Seems like it only works on real devices, not in the simulator.

lib/util/search/address_suggestion_manager.dart
```diff
@@ -97,29 +101,29 @@ class AddressSuggestionManager {
...
       "content_type": contentType,
-      "content": {"input": query}
+      "content": <String, dynamic>{"input": query}
     })));
 
-    String response = await request.close().then((value) => value.transform(utf8.decoder).join());
+    String response = await request.close().then((HttpClientResponse value) => value.transform(utf8.decoder).join());
 
     Map<String, dynamic> responseMap = json.decode(response);
     Map<String, dynamic> content = responseMap['content'];
```
istzen, line 129: line 131 is the same Exception thrown when i hit enter and the AddresSuggestions are not loaded

fynsta: I think I fixed that? Can you verify? I added a new package that helps with picking those deeply nested values. But I have no clue why it does not work with normal dart code... very confusing

lib/util/search/address_search_delegate.dart
```diff
@@ -38,7 +38,7 @@ class AddressSearchDelegate extends SearchDelegate<AddressSuggestion?> {
   // [bool mounted = true] is a hack to be able to use context
   // (StatelessWidget is always mounted, so this is fine)
   void returnFirstResult(BuildContext context, [bool mounted = true]) async {
-    final List<AddressSuggestion> suggestions = await addressSuggestionManager.getAddressSuggestions(query);
+    final List<AddressSuggestion> suggestions = await addressSuggestionManager.getSuggestions(query);
```
fynsta, line 41: fyi this is not immediately related, but I noticed that I forgot to change that when adding the other suggestion types. Of course we should return the overall first result, not the first address.

lib/util/trip/trip_card.dart
```diff
@@ -11,7 +11,7 @@ abstract class TripCard<T extends Trip> extends StatefulWidget {
   const TripCard(this.trip, {super.key});
 }
 
-abstract class TripCardState<T extends TripCard> extends State<T> {
+abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends State<U> {
```
fynsta, line 14: fyi @Showcas this is the right way to generic this, I think

lib/drives/models/drive.dart
```diff
@@ -72,17 +73,14 @@ class Drive extends Trip {
...
-    return Drive.fromJsonList(await SupabaseManager.supabaseClient.from('drives').select().eq('driver_id', userId));
-  }
+  List<Ride>? get approvedRides => rides?.where((Ride ride) => ride.status == RideStatus.approved).toList();
+  List<Ride>? get pendingRides => rides?.where((Ride ride) => ride.status == RideStatus.pending).toList();
 
   static Future<bool> userHasDriveAtTimeRange(DateTimeRange range, int userId) async {
-    //get all upcoming drives of user
-    List<Drive> drives = await Drive.getDrivesOfUser(userId);
-    drives = drives.where((drive) => !drive.cancelled && !drive.isFinished).toList();
+    List<Map<String, dynamic>> data =
```
Eddie-42, line 80: Beim Drive erstellen:
```
E/flutter ( 4364): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: type 'List<dynamic>' is not a subtype of type 'List<Map<String, dynamic>>'
E/flutter ( 4364): #0      Drive.userHasDriveAtTimeRange (package:motis_mitfahr_app/drives/models/drive.dart:80:32)
E/flutter ( 4364): <asynchronous suspension>
E/flutter ( 4364): #1      _CreateDriveFormState._onSubmit (package:motis_mitfahr_app/drives/pages/create_drive_page.dart:108:13)
E/flutter ( 4364): <asynchronous suspension>
E/flutter ( 4364):
```

fynsta: Thanks! I have now searched by hand for all occurences of `List<Map<String, dynamic>>`, I should now have actually found all of them.

Eddie-42 (approved):

Showcas (approved):

PR #67, opened 2023-01-18
Widget Tests!

fynsta:
This adds a widget test for DriveDetail. It covers 99.5% of lines (the only line not covered is the onTap of the riders that will lead to the chat after Philipp merges his PR). You will note that we had to add a few Key properties to uniquely identify the elements of the screen.

We had to refactor the test server mocking a little bit, because flutter does not allow any HTTP requests for Widget tests (not even localhost ones such as those we had before). Thus we now give a fake client directly to the supabase client. There might be more refactorings to come there, for example I would love to be able to search by Regex, and to check the body of the request.

It also adds missing relation export to `toJsonForApi` and moves some fields into "trip.toJson" which we noticed while writing the tests.

Notes: 
- The `flutter run <test-file>` functionality I talked about in the last meeting looked great at first. But after trying it out a little bit, it just seems hopelessly broken. Just run it the normal way instead.
- We removed avatarUrls for now, the mocking just did not work at all. That's a problem for the one who will do profile page tests.

co-authored by @Eddie-42


fynsta:
And I nearly forgot: We fixed a bug that happened when the drive was passed via id. The chatButton should not try to count messages in that case. @istzen this will be fun to merge with your chat PR...

lib/drives/pages/drive_detail_page.dart
```diff
@@ -369,7 +387,9 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
 
   void _cancelDrive() async {
     await _drive?.cancel();
-    setState(() {});
+    setState(() {
```
istzen, line 390: shouldn't the cancel method set cancelled true?

lib/util/buttons/custom_banner.dart
```diff
@@ -8,7 +8,8 @@ class CustomBanner extends StatelessWidget {
 
   const CustomBanner(this.text, {super.key, this.displayColorKind = DisplayColorKind.primary});
 
-  factory CustomBanner.error(text) => CustomBanner(text, displayColorKind: DisplayColorKind.error);
+  factory CustomBanner.error(text, {Key? key}) =>
+      CustomBanner(text, displayColorKind: DisplayColorKind.error, key: key);
   factory CustomBanner.warning(text) => CustomBanner(text, displayColorKind: DisplayColorKind.warning);
```
istzen, line 13: why no key here?

lib/drives/pages/drive_detail_page.dart
fynsta: yes sorry that should not have been in there. nice catch!

lib/util/buttons/custom_banner.dart
fynsta: Not needed yet... If you want I can add it for consistency though :)

test/models/profile_test.dart
```diff
@@ -173,4 +169,14 @@ void main() {
       );
     });
   });
+
+  // group('GenderName', () {
```
istzen, line 173: why is this in here?

test/widgets/drive_detail_page_test.dart
```diff
@@ -0,0 +1,164 @@
...
+      await tester.pump();
+
+      final Finder cancelDriveButton = find.byKey(const Key('cancelDriveButton'));
+      expect(cancelDriveButton, findsOneWidget);
+      await tester.scrollUntilVisible(cancelDriveButton, 500.0);
+      await tester.tap(cancelDriveButton);
+      await tester.pumpAndSettle();
+
+      final Finder cancelDriveNoButton = find.byKey(const Key('cancelDriveNoButton'));
+      expect(cancelDriveNoButton, findsOneWidget);
```
istzen, line 141: shouldn't we also check somewhere if the message in the Dialog is there?

test/widgets/drive_detail_page_test.dart
```diff
@@ -0,0 +1,164 @@
...
+      await tester.tap(cancelDriveNoButton);
+      await tester.pumpAndSettle();
+
+      verifyNever(processor.processUrl('/rest/v1/drives?id=eq.${drive.id}'));
+
+      expect(find.byKey(const Key("cancelledDriveBanner")), findsNothing);
+    });
+
+    testWidgets('can navigate to drive chat page', (WidgetTester tester) async {
```
istzen, line 150: shoudn't we also check if the chat button on the waypoints work?

test/widgets/drive_detail_page_test.dart
```diff
@@ -0,0 +1,164 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:motis_mitfahr_app/rides/models/ride.dart';
+
+import '../util/factories/drive_factory.dart';
+import '../util/factories/ride_factory.dart';
+import '../util/mock_server.dart';
+import '../util/mock_server.mocks.dart';
+import '../util/pump_material.dart';
+
+void main() {
```
istzen, line 17: what is with the riders and pending rides, can we add a test that they are there also? 
and also that a ride that is cancelledbyRider does not appear?

test/widgets/drive_detail_page_test.dart
fynsta: It does not work currently, there is a TODO there... that's why I said we can only get 100% after we merge with your PR :D

test/widgets/drive_detail_page_test.dart
fynsta: Tbh there are a million things we could check, but if there is Yes/No then we can be pretty sure that the dialog is working correctly...

test/widgets/drive_detail_page_test.dart
fynsta: Yes that would probably be a good idea 👍

Eddie-42 (approved):

Showcas (approved):

MarkFalke (approved):

fynsta:
I have refactored the tests a little bit, so there is not so much duplicate code anymore. I hope they provide a good starting point for the rest of our widget tests.

PR #69, opened 2023-01-20
Fix address suggestions scoped by account

fynsta:
The suggestions were already scoped in local storage, but they were still present in memory after logout/login. Just reloading them from the correctly scoped storage does the trick.

### DIFF
--- a/lib/util/search/address_suggestion_manager.dart
+++ b/lib/util/search/address_suggestion_manager.dart
@@ -17,8 +17,7 @@ class AddressSuggestionManager {
   static const String _storageKey = 'suggestions';
   static const int _maxSuggestionsStored = 100;
 
-  static late List<AddressSuggestion> _historySuggestions;
-  static bool _historySuggestionsLoaded = false;
+  static List<AddressSuggestion> _historySuggestions = [];
 
   // If the search is shorter than this, the server will return an empty list anyways.
   static const int searchLengthRequirement = 3;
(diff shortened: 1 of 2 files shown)


Showcas (approved):

Eddie-42 (approved):

lib/util/search/address_suggestion_manager.dart
```diff
@@ -49,6 +50,16 @@ class AddressSuggestionManager {
     Iterable<Map<String, dynamic>> suggestions = data.map((suggestion) => jsonDecode(suggestion));
     _historySuggestions =
         suggestions.map((suggestion) => AddressSuggestion.fromJson(suggestion, fromHistory: true)).toList();
+
+    _historySuggestionsLoaded = true;
+  }
+
+  void reloadHistorySuggestions() {
+    if (_historySuggestionsLoaded) {
```
Eddie-42, line 58: Is this necessary? Does this throw an error if `_historySuggestions` is empty?

fynsta: yes because it's late... of course we could make the field nullable instead, but that's not really what this is for

Eddie-42: I would've suggested to initialize it empty, I think that would be functionally equal wouldn't it? Because then you can just call clear without worrying. I understand that semantically you could separate a "real" empty list from one that is not yet loaded, but this is the only point where that case is relevant I think

fynsta: The thing is: When you open the app, the suggestions are filled with the correct value immediately. Isn't that exactly what `late` was made for?

Eddie-42: Mh okay, I guess, I don't know what the convention would be here, we can leave it

fynsta: I'll change it :D Less code

PR #70, opened 2023-01-20
More rules

fynsta:
This should hopefully be the last gigantic one of these. This enables basically all rules that the flutter repo itself has enabled (https://github.com/flutter/flutter/blob/master/analysis_options.yaml).

You can look up what those rules do here: https://dart-lang.github.io/linter/lints/. A few of these were already enabled as you have surely seen (avoid_print, prefer_const_constructors et al) by the flutter recommended lints. You can see those here: https://github.com/flutter/packages/blob/main/packages/flutter_lints/lib/flutter.yaml

The most important ones:
-  prefer_final_fields/prefer_final_in_for_each/prefer_final_locals: Exactly what those are saying, make variables final when possible
- require_trailing_commas: My favorite rule! Adds trailing commas if the line is too long otherwise. This corresponds exactly with what we are trying to do, but now the editor can help us!
- prefer_single_quotes: We could also use double quotes, but this is just to have a consistent quoting style. Double quotes are allowed in the special case that the string itself contains a single quote
- avoid_void_async/unawaited_futures/await_only_futures: Improving our handling of futures in general

Most of the others are very minor and/or currently did not create any offences in our project. I have left a few for us to enable in the future and to discuss whether we want those. PLEASE WEIGH IN (as opposed to just approving).


analysis_options.yaml
```diff
@@ -5,13 +5,169 @@
...
+    - implementation_imports
+    - iterable_contains_unrelated_type
+    - leading_newlines_in_multiline_strings
+    - library_names
+    - library_prefixes
+    - library_private_types_in_public_api
+    - list_remove_unrelated_type
+    - missing_whitespace_between_adjacent_strings
+    - no_adjacent_strings_in_list
+    # - no_default_cases # DO WE WANT THIS?
```
Showcas, line 73: I think defaults are usefull sometimes but Im ok with both

analysis_options.yaml
```diff
@@ -5,13 +5,169 @@
...
+    - prefer_void_to_null
+    - provide_deprecation_message
+    - recursive_getters
+    - require_trailing_commas
+    - secure_pubspec_urls
+    - sized_box_for_whitespace
+    - sized_box_shrink_expand
+    - slash_for_doc_comments
+    - sort_child_properties_last
+    # - sort_pub_dependencies # DO WE WANT THIS?
```
Showcas, line 128: I think its a nice feature for debuging later

lib/util/profiles/reviews/custom_rating_bar.dart
```diff
@@ -24,9 +24,9 @@ class CustomRatingBar extends StatelessWidget {
         Icons.star,
         color: Colors.amber,
       ),
+      // ignore: avoid_redundant_argument_values
```
Showcas, line 27: Will the autogenerated documentation stay in for the long run or is it just generated to explain the deletion of direction: ?

analysis_options.yaml
Eddie-42: I don't want this. I think this would be more on the functionality and less on the maintainability side because it has the potential to make a lot of code unreadable. In our case, `isMutuallyExclusive` would be problematic with this rule

analysis_options.yaml
Eddie-42: I think the problem with this is that it would clash with the current sorting, which includes some smart categorizing. If that is the case, I don't want this.

analysis_options.yaml
fynsta: Yes that was my thought. But in that case we could of course put "ignore: no_default_cases" there. But I agree that it can be a bit annoying. We should still try to follow the rule whenever possible, but I can remove it from the list

analysis_options.yaml
fynsta: Yes I wanted to look that up, let me check if we can give it some pre-defined groups for example.

lib/util/profiles/reviews/custom_rating_bar.dart
fynsta: that's not autogenerated, I put that there. Just because I like "itemCount: Review.maxRating". It's more explicit and will not break everything if we decide to change our review system to 3 stars for example.

But I will add a comment to make that more clear

Eddie-42 (approved):

analysis_options.yaml
fynsta: Okay it's impossible, I'll remove the rule

Showcas (approved):

PR #71, opened 2023-01-20
Fix capitalization of "No rides" text

fynsta:
Just a new text for every combination of ride/drive and past/upcoming. Along the way, I refactored the TripPageBuilder a little bit. It had way too many unnecessary arguments. I still think this is not perfect because there are some random type checks in there, but it's definitely a lot easier to read.

Also moves the logic for showing rides/drives in the list to the respective models. (I made it so that past pending rides are not shown anymore either, does everyone agree with that?)

Also the page is now a stateless widget, I don't know why it was just a class with a static build method?

### DIFF
--- a/lib/l10n/app_de.arb
+++ b/lib/l10n/app_de.arb
@@ -259,8 +259,10 @@
 
   "widgetTripBuilderTabUpcoming": "Zukünftig",
   "widgetTripBuilderTabPast": "Vergangen",
-  "widgetTripBuilderNoUpcoming": "Keine zukünftigen {name}.",
-  "widgetTripBuilderNoPast": "Keine vergangenen {name}.",
+  "widgetTripBuilderNoUpcomingDrives": "Keine zukünftigen Fahrten.",
+  "widgetTripBuilderNoUpcomingRides": "Keine zukünftigen Mitfahrten.",
+  "widgetTripBuilderNoPastDrives": "Keine vergangenen Fahrten.",
+  "widgetTripBuilderNoPastRides": "Keine vergangenen Mitfahrten.",
 
   "pageSearchRideTitle": "Mitfahrt suchen",
   "pageSearchRideButtonSearch": "Suchen",
--- a/lib/l10n/app_de.arb
+++ b/lib/l10n/app_de.arb
@@ -259,10 +259,10 @@
 
   "widgetTripBuilderTabUpcoming": "Zukünftig",
   "widgetTripBuilderTabPast": "Vergangen",
-  "widgetTripBuilderNoUpcomingDrives": "Keine zukünftigen Fahrten.",
-  "widgetTripBuilderNoUpcomingRides": "Keine zukünftigen Mitfahrten.",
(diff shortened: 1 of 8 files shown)


lib/l10n/app_de.arb
```diff
@@ -259,8 +259,10 @@
 
   "widgetTripBuilderTabUpcoming": "Zukünftig",
   "widgetTripBuilderTabPast": "Vergangen",
-  "widgetTripBuilderNoUpcoming": "Keine zukünftigen {name}.",
-  "widgetTripBuilderNoPast": "Keine vergangenen {name}.",
+  "widgetTripBuilderNoUpcomingDrives": "Keine zukünftigen Fahrten.",
```
Showcas, line 262: I would remove the '.' at the end. We put the '.' for sentences but I would not use them for hints or small infos

Showcas: If you would do 'Du hast keine ...' then the '.' can be there in my opinion

lib/rides/models/ride.dart
```diff
@@ -126,6 +126,12 @@ class Ride extends Trip {
     return false;
   }
 
+  @override
+  bool shouldShowInListView({required bool past}) {
+    return super.shouldShowInListView(past: past) &&
+        (!past || (status != RideStatus.pending && status != RideStatus.withdrawnByRider));
```
Showcas, line 132: RideStatus.withdrawnByRider implies hideInListView. Since it should change both when done. If we say we still leave it in for safty then I would recommend to also check for RideStatus.Preview

Showcas:
Rest looks good to me

lib/util/trip/trip_page_builder.dart
```diff
@@ -38,31 +35,47 @@ class TripPageBuilder {
...
+        floatingActionButton: Semantics(
+          sortKey: const OrdinalSortKey(0),
+          child: getFloatingActionButton(isRide, context),
+        ),
       ),
     );
   }
+
+  Widget getFloatingActionButton(bool isRide, BuildContext context) {
+    if (isRide) {
```
Eddie-42, line 63: I think it would be better if you used the `?` operator or variables. `onPressed` and `backgroundColor` are still the same, and I think it should be visible that this is logically the same button

lib/rides/models/ride.dart
fynsta: Yes I know, I just wanted to make this explicit. I could add preview here, but it would be confusing if you consider the Search-"ListView" where it is shown.

Eddie-42 (approved):

lib/l10n/app_de.arb
fynsta: I removed the dot.

@Eddie-42 shrug would be even better, right?

lib/util/trip/trip_page_builder.dart
fynsta: I thought about that too. It's ride on the edge where 3 of 5 arguments are different lol.

Showcas (approved):

MarkFalke (approved):

PR #72, opened 2023-01-21
Set to system locale on first start

fynsta:
Fixes the app language on first startup. It should be read be from the system language. We will ignore any further change in the system language (should we do that?).

### DIFF
--- a/lib/util/locale_manager.dart
+++ b/lib/util/locale_manager.dart
@@ -1,6 +1,5 @@
 import 'dart:io';
 
-import 'package:collection/collection.dart';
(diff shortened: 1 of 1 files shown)


Eddie-42 (approved):
LGTM, although I'm pretty sure we can only test this when we install the next APK

fynsta:
> LGTM, although I'm pretty sure we can only test this when we install the next APK

No, you can also change the language of the emulator and then do `flutter run`. It's the same thing

Showcas (approved):

PR #73, opened 2023-01-21
Better create drive

Eddie-42:
As we talked about, I redesigned the create drive page a bit (but without changing the functionality). A similar update to the search rides page is coming later.

### DIFF
--- a/lib/drives/pages/create_drive_page.dart
+++ b/lib/drives/pages/create_drive_page.dart
@@ -5,9 +5,10 @@ import 'package:supabase_flutter/supabase_flutter.dart';
 import '../../account/models/profile.dart';
 import '../../rides/models/ride.dart';
 import '../../util/buttons/button.dart';
+import '../../util/fields/increment_field.dart';
 import '../../util/locale_manager.dart';
-import '../../util/search/address_search_field.dart';
 import '../../util/search/address_suggestion.dart';
+import '../../util/search/start_destination_timeline.dart';
 import '../../util/supabase.dart';
 import '../models/drive.dart';
 import '../pages/drive_detail_page.dart';
@@ -57,6 +58,22 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
   late int _dropdownValue;
 
   final List<int> list = List<int>.generate(10, (int index) => index + 1);
+
+  @override
+  void initState() {
+    super.initState();
+    _selectedDate = DateTime.now();
+    _dropdownValue = list.first;
+  }
+
+  @override
+  void dispose() {
+    _dateController.dispose();
+    _timeController.dispose();
+    _startController.dispose();
+    _destinationController.dispose();
+    super.dispose();
+  }
 
   void _showTimePicker() {
     showTimePicker(
@@ -174,22 +191,6 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
   }
 
   @override
-  void initState() {
-    super.initState();
-    _selectedDate = DateTime.now();
-    _dropdownValue = list.first;
-  }
-
-  @override
-  void dispose() {
-    _dateController.dispose();
-    _timeController.dispose();
-    _startController.dispose();
-    _destinationController.dispose();
-    super.dispose();
-  }
-
-  @override
   Widget build(BuildContext context) {
     // This needs to happen on rebuild to make sure we pick up locale changes
     _dateController.text = localeManager.formatDate(_selectedDate);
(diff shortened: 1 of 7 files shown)


lib/drives/pages/create_drive_page.dart
```diff
@@ -193,78 +194,65 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
...
                 ),
-              ],
+              ),
+            ],
+          ),
+          const Divider(),
+          SizedBox(
+            width: 150,
+            child: IncrementField(
+              maxValue: 8,
```
fynsta, line 243: let's set that max value inside the class instead

fynsta: inside `Trip` I guess

lib/util/fields/increment_field.dart
```diff
@@ -0,0 +1,72 @@
...
+
+class IncrementField extends StatefulWidget {
+  final int minValue;
+  final int maxValue;
+  final int initialValue;
+  final Widget? icon;
+  final void Function(int?)? onChanged;
+
+  const IncrementField(
+      {super.key, this.minValue = 1, this.initialValue = 1, this.maxValue = 10, this.icon, this.onChanged});
```
fynsta, line 11: trailing comma

lib/drives/pages/create_drive_page.dart
```diff
@@ -193,78 +194,65 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
...
+              icon: Icon(
+                Icons.chair,
+                color: Theme.of(context).colorScheme.primary,
+              ),
+              onChanged: (int? value) {
+                setState(() {
+                  _dropdownValue = value!;
```
fynsta, line 250: dropdown value does not make much sense anymore, and is also not descriptive of what the field is for

lib/util/fields/increment_field.dart
```diff
@@ -0,0 +1,72 @@
+import 'package:flutter/material.dart';
+
+class IncrementField extends StatefulWidget {
+  final int minValue;
+  final int maxValue;
+  final int initialValue;
+  final Widget? icon;
+  final void Function(int?)? onChanged;
```
fynsta, line 8: why is the function nullable?

lib/util/search/address_search_field.dart
```diff
@@ -7,18 +7,18 @@ import 'address_suggestion.dart';
 class AddressSearchField extends StatelessWidget {
   final AddressType addressType;
   final TextEditingController controller;
-  final void Function(AddressSuggestion) onSelected;
+  final void Function(AddressSuggestion)? onSelected;
```
fynsta, line 10: I don't understand this change? seems unneeded

fynsta:
Definitely like the new design!

lib/drives/pages/create_drive_page.dart
Eddie-42: Yeah I know, it's already fixed

lib/util/search/address_search_field.dart
Eddie-42: Seeing this as a general class that could be used outside of our context with suggestion objects that have to be set, I see no reason why you would have to provide a function here. In most cases, a controller would suffice

fynsta: I'm sorry, but I don't agree - the whole thing does not make sense if we don't have `AddressSuggestion` objects - it's literally what the `SearchField` was made for

Showcas: Is there a case where we would need it without an object?

Eddie-42: Why not give users the choice? If I just want a "Choose your favorite place in the world" textfield and just wanted to store the string, I feel like I should be able to do that without needing to provide a (){} function

fynsta: We are not writing a library for other people, we are writing an app the we use ourselves...

Showcas (approved):

lib/util/search/start_destination_timeline.dart
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return FixedTimeline(
+      theme: CustomTimelineTheme.of(context),
+      children: <Widget>[
+        TimelineTile(
+          contents: Padding(
+            padding: const EdgeInsets.all(4.0),
```
istzen, line 29: can we add a little bit more Padding to the left. "Das braucht Platz zum Atmen" zitat by Merri

lib/util/trip/trip.dart
```diff
@@ -4,6 +4,8 @@ import '../model.dart';
 import '../search/position.dart';
 
 abstract class Trip extends Model {
+  static const int maxSelectableSeats = 8;
```
istzen, line 7: why did we reduce that to 8?

lib/drives/pages/create_drive_page.dart
```diff
@@ -199,78 +199,66 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
...
+              ),
+            ],
+          ),
+          const Divider(),
+          SizedBox(
+            width: 150,
+            child: IncrementField(
+              initialValue: _seats,
+              maxValue: Trip.maxSelectableSeats,
+              icon: Icon(
```
istzen, line 250: can we make the chair icon a little bit bigger

lib/drives/pages/create_drive_page.dart
```diff
@@ -199,78 +199,66 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
...
                   ),
                 ),
-              ],
+              ),
+            ],
+          ),
+          const Divider(),
+          SizedBox(
+            width: 150,
+            child: IncrementField(
```
istzen, line 247: can we maybe add some kind of tooltip or also a Border around that so it's clear that you are selecting the seats.

lib/util/trip/trip.dart
istzen: i'll change it in the database so it's only >= 1

fynsta: Oh okay then that should be changed in this PR as well.

Although that makes the display more difficult for us... The 100 seat icons in a row would overflow completely.

Eddie-42: I think we should place some sort of limit on the number of seats. If your cat taps on the + icon 100 times, you have to click on the - icon 100 times or I have to implement a reset functionality. I think the increment field is not really designed to handle even high 2-digit numbers and at some point we should safeguard against silly inputs as well. What Fynn found isn't a car anymore, it's under the tab "trucks and vans" on the website, at which point we have to decide whether we want to accomodate 49-seater buses as well. BlaBlaCar's limit is 4, by the way!

fynsta (approved):

lib/util/trip/trip.dart
istzen: then let's just leave 8?

lib/util/fields/increment_field.dart
```diff
@@ -31,48 +32,52 @@ class _IncrementFieldState extends State<IncrementField> {
...
-            alignment: Alignment.centerLeft,
-            child: IconButton(
-              icon: const Icon(Icons.remove_circle_outline),
-              onPressed: _value <= widget.minValue
-                  ? null
-                  : () {
-                      _value -= 1;
-                      widget.onChanged(_value);
-                    },
+    return Card(
```
istzen, line 35: Maybe not a card, but a border like for the other Inputs? and for people who don't get that that should be the seats write seats somewhere (Merri did not get it right away)

fynsta: Hm everybody in the user study got that part right away. Should probably wait for some more feedback there?

lib/util/trip/trip.dart
fynsta: Yes, can you add it in supabase again?

fynsta:
@istzen Can we merge this from your POV? Making the seats clearer can be done after finishing up the user studies.

istzen (approved):

PR #74, opened 2023-01-21
Fix coverage percentage

fynsta:
Because it only includes files that are actually loaded. This adds a useful shell script for generating a file that just randomly imports all other files... That file will also be run in the pipeline.

More info on the flutter issue: https://github.com/flutter/flutter/issues/27997


scripts/coverage_helper.sh
```diff
@@ -0,0 +1,12 @@
...
+
+file=test/coverage_helper_test.dart
+printf "// Helper file to make coverage work for all dart files\n" > $file
+printf "// **************************************************************************\n" >> $file
+printf "// Because of this: https://github.com/flutter/flutter/issues/27997#issue-410722816\n" >> $file
+printf "// DO NOT EDIT THIS FILE USE: ./scripts/file_finder.sh\n" >> $file
+printf "// **************************************************************************\n" >> $file
+printf "\n" >> $file
+printf "// ignore_for_file: unused_import\n" >> $file
+find lib -type f \( -iname "*.dart" ! -iname "*.g.dart" ! -iname "*.freezed.dart" ! -iname "generated_plugin_registrant.dart" \) | cut -c4- | awk -v package="$1" '{printf "import '\''package:motis_mitfahr_app%s'\'';\n", $1}' >> $file
```
Eddie-42, line 11: Yes I can definitely understand and am qualified to review this

fynsta: Basically it finds all files that end with ".dart", but not ".g.dart" etc. And then it creates an import line for each of those file names.

You can just run the script and see the file that it results in.

Showcas (approved):

Eddie-42 (approved):

Eddie-42:
Does this also impact the comments by CodeCov/How does that work?

fynsta:
> Does this also impact the comments by CodeCov/How does that work?

yes it does. Codecov gets sent the results of the test run in the pipeline. Since I put that shell script in the pipeline immediately before that step, this will be applied.

PR #75, opened 2023-01-21
Better search ride

Eddie-42:
This overhauls the system with which you can search for rides:
- The search suggestion page has been renamed to search ride page, and you will get directly to the former search suggestion page if you click on the Search FAB. This is because every search parameter is present on that page.
- There is now an option to search for rides on a whole day without providing a time, which is turned on by default. If you turn it off, you will see an additional time input. Furthermore, the filter category "Max deviation from selected time" will be given in days or hours, respectively.
- You can now switch your start and destination easily with a single button next to the input fields.
- If you scroll through the page, the AppBar will now disappear along with the time, seats and filters, leaving only the start and destination at the top. This is because while testing on the device, we noticed that the page was completely unusable in landscape or split screen mode because the options at the top of the page were so big that you couldn't see a single suggestion.

A few smaller changes: Most importantly, I had to give the avatar a parameter if heros should be displayed or not because otherwise, multiple heros with the same key would be displayed at the same time sometimes

Sorry that this isn't based on the main branch, I had to reuse a few widgets that I had already build for the create drive page in order to achieve a coherent design.


lib/account/widgets/avatar.dart
```diff
@@ -12,36 +12,40 @@ class Avatar extends StatefulWidget {
...
+      radius: widget.size,
+      backgroundImage: widget.profile.avatarUrl?.isEmpty ?? true ? null : NetworkImage(widget.profile.avatarUrl!),
+      child: widget.profile.avatarUrl?.isEmpty ?? true
+          ? Text(
+              widget.profile.username[0].toUpperCase(),
+              style: TextStyle(fontSize: widget.size),
+            )
+          : null,
+    );
+    return widget.withHero ? Hero(tag: "Avatar-${widget.profile.id}", child: avatar) : avatar;
```
fynsta, line 41: quick question because that is hard to see from the diff: Which avatars do not need the hero?

lib/l10n/app_de.arb
```diff
@@ -263,26 +263,25 @@
...
-  "searchSuggestionsFilterDeviationHours": "Stunden",
-  "searchSuggestionsFilterSorting": "Sortierung",
-  "searchSuggestionsFilterResetToDefault": "Zurücksetzen",
-  "searchSuggestionsSortingRelevance": "Relevanz",
-  "searchSuggestionsSortingTimeProximity": "Zeitliche Nähe",
-  "searchSuggestionsSortingTravelDuration": "Reisedauer",
-  "searchSuggestionsSortingPrice": "Preis",
+  "pageSearchRideNoInput": "Gib dein Start und Ziel ein",
+  "pageSearchRideEmpty": "Nichts zu sehen",
+  "pageSearchRideRelaxRestrictions": "Bitte locker deine Suchbeschränkungen.",
```
fynsta, line 268: "lockere"

fynsta:
Can you explain for all of us what the new package does?

lib/rides/pages/search_ride_page.dart
```diff
@@ -125,133 +105,270 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+      return SliverList(
+        delegate: SliverChildBuilderDelegate((BuildContext context, int index) {
+          final Ride ride = filteredSuggestions[index];
+          return RideCard(ride);
+        }, childCount: filteredSuggestions.length),
+      );
+    }
+  }
+
+  Widget buildDateRow() {
```
Showcas, line 283: I love the lokk and the way you set it up. I don't feels so good about The Date, Seats and Filter being attached to the Scrolleable at the top. Couldn't we like make it a DropDown and hide it the moment we start scrolling?

lib/rides/widgets/search_ride_filter.dart
```diff
@@ -197,7 +202,7 @@ class SearchSuggestionFilter {
   Widget _buildDeviationFilter(BuildContext context) {
```
fynsta, line 202: while we're at it, can we make this a better input type? It should be a field with up and down buttons for whole numbers (similar to the seat input, but of course different design)

lib/rides/pages/search_ride_page.dart
```diff
@@ -125,133 +105,270 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+            ],
+          ),
+        ),
+      );
+    } else {
+      return SliverList(
+        delegate: SliverChildBuilderDelegate((BuildContext context, int index) {
+          final Ride ride = filteredSuggestions[index];
+          return RideCard(ride);
+        }, childCount: filteredSuggestions.length),
```
fynsta, line 278: trailing comma missing

lib/rides/pages/search_ride_page.dart
```diff
@@ -125,133 +105,270 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+          const SizedBox(width: 10),
+        ],
+        Expanded(
+          child: Row(
+            children: <Widget>[
+              Checkbox(
+                value: _wholeDay,
+                onChanged: (bool? value) => setState(() => _wholeDay = value!),
+              ),
+              const Text("Whole day")
```
fynsta, line 300: translation missing

lib/rides/pages/search_ride_page.dart
```diff
@@ -125,133 +105,270 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
 
-  Widget buildSeatsPicker() {
-    return Expanded(
+  Widget buildMainContentSliver() {
+    if (_rideSuggestions == null || _loading) {
+      if (_startSuggestion == null || _destinationSuggestion == null) {
+        return SliverToBoxAdapter(
+          child: Column(
+            children: <Widget>[
+              Image.asset('assets/pointing_up.png'),
```
Showcas, line 218: I think its better without the Emojis. Just the nice test telling you what to do in the center

lib/rides/pages/search_ride_page.dart
```diff
@@ -125,133 +105,270 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+                    floating: true,
+                    flexibleSpace: FlexibleSpaceBar(
+                      background: Container(color: Theme.of(context).colorScheme.surface),
+                      title: Text(
+                        S.of(context).pageSearchRideTitle,
+                        style: TextStyle(color: Theme.of(context).colorScheme.onSurface),
+                      ),
+                    ),
+                  ),
+                  SliverPinnedHeader(
```
Showcas, line 351: Could we also pin the Appbar, just feel like you should always be able to click on back from any position

lib/rides/widgets/search_ride_filter.dart
```diff
@@ -209,7 +214,7 @@ class SearchSuggestionFilter {
               inputFormatters: <TextInputFormatter>[FilteringTextInputFormatter.digitsOnly],
             ),
           ),
-          Text(S.of(context).searchSuggestionsFilterDeviationHours),
+          Text(wholeDay ? S.of(context).searchRideFilterDeviationDays : S.of(context).searchRideFilterDeviationHours),
```
fynsta, line 217: this does not seem to work correctly, I had switched off wholeDay and it still shows days

lib/rides/pages/search_ride_page.dart
fynsta: Also please, make the whole text clickable

lib/rides/pages/search_ride_page.dart
Eddie-42: Do you mean that it should immediately show again the moment you start scrolling up like the AppBar? I agree, but that is incredibly complicated to implement and I would like to leave it for now

lib/l10n/app_de.arb
Eddie-42: Locker ma deine Base digga

Eddie-42:
> Can you explain for all of us what the new package does?

The new package provides additional functionality for Flutter slivers, since what Flutter provides is very basic. With Flutter, for every new behavior (not just "Scrolling as usual"), you would either have to use an AppBar because for some reason that works or create at least one new class and solve some problems (I've tried that, it's a bit complicated). So this class provides a persistent header, for example, that stays at the top once it reaches it, and I used that for the location fields

lib/rides/pages/search_ride_page.dart
Eddie-42: I thought we should use little cute pictures for every one of these kind of "empty" messages, since an empty page with only text would look very bland I suppose

lib/rides/pages/search_ride_page.dart
Eddie-42: That would be very easy, but it's also very common for Android apps to do this. You can still go back via the gesture or access the app bar immediately if you start to scroll up again

lib/rides/pages/search_ride_page.dart
fynsta: I like the emoji too, it definitely fits with the style we have used on other pages

lib/rides/pages/search_ride_page.dart
Showcas: Its ok, we can add it later then if we have time

fynsta: Time? what is that? 🤣

lib/rides/widgets/search_ride_filter.dart
```diff
@@ -35,8 +33,9 @@ class SearchSuggestionFilter {
   late int _minReliabilityRating;
   late int _minHospitalityRating;
   late List<Feature> _selectedFeatures;
-  late SearchSuggestionSorting _sorting;
-  final TextEditingController _maxDeviationController = TextEditingController();
+  late SearchRideSorting _sorting;
+
+  bool _wholeDay;
```
Showcas, line 38: I don't know about the whole day thing. So we are saying we filter for rides only at that day and you don't see anything the day after?

Showcas (approved):

lib/l10n/app_de.arb
```diff
@@ -265,26 +265,26 @@
...
+  "pageSearchRideEmpty": "Nichts zu sehen",
+  "pageSearchRideRelaxRestrictions": "Bitte lockere deine Suchbeschränkungen.",
+  "pageSearchRideNoResults": "Es tut uns leid, wir konnten für diese Route keine Mitfahrgelegenheit finden.",
+  "pageSearchRideTooltipFilter": "Filter",
+  "pageSearchRideWholeDay": "Ganztägig",
+
+  "searchRideFilterMinimumRating": "Mindest-Bewertung",
+  "searchRideFilterFeatures": "Features",
+  "searchRideFilterFeaturesSelectTooltip": "Auswählen",
+  "searchRideFilterFeaturesDeselectTooltip": "Abwählen ",
```
istzen, line 278:
```suggestion
  "searchRideFilterFeaturesDeselectTooltip": "Entfernen",
```
I think "Abwählen" sounds kinda weird

lib/rides/pages/search_ride_page.dart
```diff
@@ -126,133 +103,296 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+        if (_wholeDay) ...<Widget>[
+          Row(
+            children: <Widget>[
+              IconButton(
+                onPressed: _selectedDate.isSameDayAs(DateTime.now())
+                    ? null
+                    : () => setState(() => _selectedDate = _selectedDate.subtract(const Duration(days: 1))),
+                icon: const Icon(Icons.chevron_left),
+              ),
+              Text(localeManager.formatDate(_selectedDate)),
```
istzen, line 300: can we add that, we can select the datePicker here also? i think it's weird that that's different if i select whole day or not. we could leave the chevron icon just if i tap the text that the picker shows if i want to look more than a few days in the future thats way easier with the picker.

lib/rides/pages/search_ride_page.dart
```diff
@@ -126,133 +103,296 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
-          value: _dropdownValue,
-          icon: const Icon(Icons.arrow_downward),
-          decoration: InputDecoration(
-            border: const OutlineInputBorder(),
-            labelText: S.of(context).seats,
+        width: 150,
+        child: IncrementField(
+          maxValue: Trip.maxSelectableSeats,
+          icon: Icon(
+            Icons.chair,
```
istzen, line 339: again same as in better-create-drive

lib/rides/pages/search_ride_page.dart
```diff
@@ -126,133 +103,296 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+    setState(() => _loading = true);
+    final List<Map<String, dynamic>> data = parseHelper.parseListOfMaps(
+      await SupabaseManager.supabaseClient.from('drives').select('''
+          *,
+          driver:driver_id (
+            *,
+            profile_features (*),
+            reviews_received: reviews!reviews_receiver_id_fkey(*)
+          )
+        ''').eq('start', _startController.text),
```
istzen, line 118: can we maybe add here also that the only the whole day of the datecontroller is selected? or is that another PR and we leave if for now?

lib/rides/pages/search_ride_page.dart
fynsta: Yes actually that's one thing we also already learned from the user studies. That will definitely be added.

lib/rides/pages/search_ride_page.dart
Eddie-42: I wouldn't do this in the databaes query because that would require the results to reload every time you select another day. This would of course mean some more traffic, but I think that is mitigated because we only allow users to create drives up to - what, 3 months? - in the future. And in the end, this is of course replaced by the algorithm

lib/util/buttons/labeled_checkbox.dart
```diff
@@ -0,0 +1,36 @@
...
+    return InkWell(
+      onTap: () {
+        onChanged(!value);
+      },
+      child: Row(
+        mainAxisSize: MainAxisSize.min,
+        children: <Widget>[
+          Checkbox(
+            value: value,
+            onChanged: (bool? newValue) {
```
fynsta, line 26: when is the bool null? and shouldn't we return in that case?

Eddie-42: Not my choice, the Flutter checkbox expects type `void Function(bool?)?` here. This is only relevant for tristate checkboxes, though.

lib/rides/widgets/search_ride_filter.dart
```diff
@@ -370,16 +365,6 @@ class SearchSuggestionFilter {
...
-      final Widget deviationWidget = Row(
-        children: <Widget>[
-          const Icon(Icons.schedule),
-          const SizedBox(width: 6),
-          Text('± ${_maxDeviationController.text}')
-        ],
-      );
-      widgets.add(deviationWidget);
-    }
     final Widget sortingWidget = Row(
```
fynsta, line 368: I know we wanted to make it so this is not part of the Inkwell, but instead leads to a different dropdown. Do we want to do that in a different PR after the user studies, or now?

lib/util/buttons/labeled_checkbox.dart
fynsta: Yes but the next line would throw an error in that case because of the strict null check. And I think we should maybe handle that more gracefully.

lib/rides/widgets/search_ride_filter.dart
Eddie-42: I'd do it after the user studies

fynsta (approved):

lib/rides/pages/search_ride_page.dart
istzen: yes that's right. I just thought it's kinda weird, that when you only get the rides of the day shown when, whole day is selected and not when a date is selected if i am correct

lib/rides/pages/search_ride_page.dart
```diff
@@ -126,133 +104,292 @@ class _SearchRideFormState extends State<SearchRideForm> {
...
+      _loading = false;
+    });
+  }
+
+  Widget buildSearchFieldViewer() {
+    return Row(
+      children: <Widget>[
+        Expanded(
+          child: Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 6),
```
istzen, line 149: can we add some more padding to the top.

MarkFalke (approved):

lib/rides/pages/search_ride_page.dart
Eddie-42: Yes I think the idea has been floated around to also only show drives on the same day without whole day selected, so that the time only makes your search more precise, but I'm not sure if we've discussed that properly

istzen: yes that seems like a good idea

istzen (approved):

PR #77, opened 2023-01-21
Add RideDetailPage tests

fynsta:
Second try without broken diff

### DIFF
--- a/lib/rides/models/ride.dart
+++ b/lib/rides/models/ride.dart
@@ -203,6 +203,7 @@ extension RideStatusExtension on RideStatus {
     return this == RideStatus.approved;
   }
 
+  // This is used to determine if the rider can see and be seen in the riders list
   bool isRealRider() {
     return this == RideStatus.approved || this == RideStatus.cancelledByDriver;
   }
--- a/lib/rides/models/ride.dart
+++ b/lib/rides/models/ride.dart
@@ -203,6 +203,10 @@ extension RideStatusExtension on RideStatus {
     return this == RideStatus.approved;
   }
 
+  bool isRealRider() {
+    return this == RideStatus.approved || this == RideStatus.cancelledByDriver;
+  }
+
   bool activeChat() {
     return this == RideStatus.approved || this == RideStatus.cancelledByDriver || this == RideStatus.cancelledByRider;
   }
--- a/test/widgets/ride_detail_page_test.dart
+++ b/test/widgets/ride_detail_page_test.dart
@@ -373,7 +373,7 @@ void main() {
           await tester.pumpAndSettle();
 
           // Verify that the ride was requested (but no way to verify body right now)
-          // One call to update the ride, one call to load it again
+          // One call to insert the ride, one call to load it again
           verify(processor.processUrl(argThat(startsWith('/rest/v1/rides')))).called(2);
 
           expect(find.byKey(const Key('rideRequestedBanner')), findsOneWidget);
--- a/test/widgets/ride_detail_page_test.dart
+++ b/test/widgets/ride_detail_page_test.dart
@@ -1,0 +1,484 @@
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:motis_mitfahr_app/account/models/profile.dart';
+import 'package:motis_mitfahr_app/account/pages/write_review_page.dart';
+import 'package:motis_mitfahr_app/account/widgets/features_column.dart';
+import 'package:motis_mitfahr_app/account/widgets/reviews_preview.dart';
+import 'package:motis_mitfahr_app/drives/models/drive.dart';
+import 'package:motis_mitfahr_app/rides/models/ride.dart';
+import 'package:motis_mitfahr_app/rides/pages/ride_detail_page.dart';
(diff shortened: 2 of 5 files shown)


Showcas:
please don't forget the page when you name the test. Just thought we are doing the same tests

Showcas (approved):

fynsta:
> please don't forget the page when you name the test for the PR titel. Just thought we are doing the same tests 

whoops my bad :D

lib/rides/models/ride.dart
```diff
@@ -203,6 +203,10 @@ extension RideStatusExtension on RideStatus {
     return this == RideStatus.approved;
   }
 
+  bool isRealRider() {
```
istzen, line 206: can we add a comment, what this method checks, i think the name is kinda weird

fynsta: Yes I'm not happy with the name either... Any ideas?

test/widgets/ride_detail_page_test.dart
```diff
@@ -0,0 +1,554 @@
...
+        ride = RideFactory().generateFake(
+          status: RideStatus.preview,
+          drive: NullableParameter(null),
+          driveId: drive.id,
+        );
+
+        when(processor.processUrl(any)).thenReturn(jsonEncode(drive.toJsonForApi()));
+      });
+
+      testWidgets('It loads the corresponding drive', (WidgetTester tester) async {
```
istzen, line 110: i don't understand the test description

test/widgets/ride_detail_page_test.dart
```diff
@@ -0,0 +1,554 @@
...
+    });
+
+    group('when ride is cancelled by driver', () {
+      setUp(() {
+        ride = RideFactory().generateFake(status: RideStatus.cancelledByDriver, drive: NullableParameter(drive));
+
+        when(processor.processUrl(any)).thenReturn(jsonEncode(ride.toJsonForApi()));
+      });
+
+      testWidgets('it shows the riders, a hide button and a banner', (WidgetTester tester) async {
```
istzen, line 191: why are we showing the riders here but not when the ride is cancelledbyRider?

test/widgets/ride_detail_page_test.dart
```diff
@@ -0,0 +1,554 @@
...
+          final Ride returnedRide = RideFactory().generateFake(status: RideStatus.pending);
+          when(processor.processUrl(any)).thenReturn(jsonEncode(returnedRide.toJsonForApi()));
+
+          final Finder requestRideYesButton = find.byKey(const Key('requestRideYesButton'));
+          expect(requestRideYesButton, findsOneWidget);
+          await tester.tap(requestRideYesButton);
+          await tester.pumpAndSettle();
+
+          // Verify that the ride was requested (but no way to verify body right now)
+          // One call to update the ride, one call to load it again
```
istzen, line 376:
```suggestion
          // One call to insert the ride, one call to load it again
```

lib/rides/models/ride.dart
istzen: no that's the problem xD, maybe just a short comment or something.

test/widgets/ride_detail_page_test.dart
fynsta: Yeah no idea what that was supposed to mean

test/widgets/ride_detail_page_test.dart
fynsta: well if its cancelledByDriver, then the state is basically frozen - no one can request or withdraw etc. So it's fine if show all the other riders (only overlapping of course) that were equally screwed by the cancellation.

But if it's cancelledByRider, then you are not a part of the drive anymore, so you should not see any riders that are added after you cancelled for example

lib/rides/models/ride.dart
fynsta: maybe "canSeeFullRideInformation" ?

test/widgets/ride_detail_page_test.dart
istzen: that makes sense thanks

lib/rides/models/ride.dart
istzen: yes that's good then it's clear why we need the method.

istzen (approved):

test/widgets/ride_detail_page_test.dart
```diff
@@ -0,0 +1,556 @@
...
+import 'package:motis_mitfahr_app/welcome/pages/login_page.dart';
+import 'package:motis_mitfahr_app/welcome/pages/register_page.dart';
+
+import '../util/factories/drive_factory.dart';
+import '../util/factories/model_factory.dart';
+import '../util/factories/profile_factory.dart';
+import '../util/factories/review_factory.dart';
+import '../util/factories/ride_factory.dart';
+import '../util/mock_server.dart';
+import '../util/mock_server.mocks.dart';
```
Eddie-42, line 29: Hast einen Rebase verkackt:
```
Target of URI doesn't exist: '../util/mock_server.mocks.dart'.
Try creating the file referenced by the URI, or Try using a URI for a file that does exist.
```

Eddie-42 (approved):

test/widgets/ride_detail_page_test.dart
fynsta: Jaja bin am fixen... ist nur leider bisschen mehr

PR #79, opened 2023-01-21
Add support for coverage for windows

Eddie-42:
@Showcas Workflow to install lcov for Windows is described here: https://fredgrott.medium.com/lcov-on-windows-7c58dda07080
After that, it should work. I had to make this change because the open command is called start on Windows.


fynsta (approved):

Showcas (approved):

fynsta:
@Showcas does that mean this works for you?

istzen (approved):

PR #80, opened 2023-01-23
Test trips model

Showcas:
100% codecoverage for:
- trip model,
- ride model,
- drive model

Removed unused methods in ride


test/models/drive_test.dart
```diff
@@ -0,0 +1,1175 @@
...
+// Um herauszufinden welche URL durch die jeweilige Datenbankabfrage generiert wird, einfach den auskommentierten Print-Aufruf in der mockServer.Dart Datei aktivieren
+
+void main() {
+  final MockUrlProcessor driveProcessor = MockUrlProcessor();
+  //setup muss in jeder Testklasse einmal aufgerufen werden
+  setUp(() async {
+    MockServer.setProcessor(driveProcessor);
+  });
+  group('create a Drive', () {
+    test('The Types are correct for a Drive', () {
```
fynsta, line 30: I don't think that's a useful test, how could those types be wrong?

test/models/drive_test.dart
```diff
@@ -0,0 +1,1175 @@
...
+        createDependencies: false,
+      );
+      expect(
+        drive.toString(),
+        'Drive{id: 1, from: start at 2022-02-02 00:00:00.000Z, to: end at 2023-03-03 00:00:00.000Z, by: 5}',
+      );
+    });
+  });
+  //soll equals auch auf rides prüfen und auf driver?
+  group('Drive.equals', () {
```
fynsta, line 791: Do we want to write these tests for every model... I feel like that's just a few hundred lines of code that is not particularly insightful

test/models/drive_test.dart
```diff
@@ -0,0 +1,1175 @@
...
+        ],
+      );
+      expect(drive.approvedRides, []);
+    });
+    test('return a List with 1 ride if rides has rides only has 1 ride and it is approved', () async {
+      final Ride ride = RideFactory().generateFake(status: RideStatus.approved);
+      final Drive drive = DriveFactory().generateFake(rides: [ride]);
+      expect(drive.approvedRides, [ride]);
+    });
+    test('return the right ride with a rides have 1 approved in it', () async {
```
fynsta, line 200: can you check the language here and everywhere else? this is just not proper grammar and I'm not even sure what it's trying to say

fynsta:
can you please stop using the normal "git pull" to merge the branches (or whatever you are doing, I'm not completely sure...). The commit history looks pretty broken, and I can't rebase this branch on main now.

Showcas:
> can you please stop using the normal "git pull" to merge the branches (or whatever you are doing, I'm not completely sure...). The commit history looks pretty broken, and I can't rebase this branch on main now.

I used the normal pull and rebase

test/models/drive_test.dart
Showcas: ahh yes sorry

test/models/drive_test.dart
Showcas: ok

fynsta:
> I used the normal pull to get main and then I use rebase. how else am I suppose to do it? merge hab ich seit deiner letzten bitte nicht mehr benutzt

Oh okay then I'm sorry. I'm confused why the git rebase behaviour is broken in that case... And this commit looks suspiciously like a merge commit: https://github.com/MOTIS-Mitfahr-App/flutter-app/pull/80/commits/828345150f6dc6f4cc449f13187fce1de9c96f6e

fynsta:
And I think it's also why Codecov does not work for this PR. It cannot find a base commit. Not sure what happened

Showcas:
> > I used the normal pull to get main and then I use rebase. how else am I suppose to do it? merge hab ich seit deiner letzten bitte nicht mehr benutzt
> 
> Oh okay then I'm sorry. I'm confused why the git rebase behaviour is broken in that case... And this commit looks suspiciously like a merge commit: [8283451](https://github.com/MOTIS-Mitfahr-App/flutter-app/commit/828345150f6dc6f4cc449f13187fce1de9c96f6e)

I noticed it aswell mhhh. We can talk later about it maybe Im doing something wrong

test/models/ride_test.dart
```diff
@@ -0,0 +1,961 @@
...
+          await Ride.userHasRideAtTimeRange(
+              DateTimeRange(
+                start: DateTime.now().add(const Duration(hours: 4)),
+                end: DateTime.now().add(const Duration(hours: 15)),
+              ),
+              2),
+          false);
+    });
+  });
+  group('Ride.shouldShowInListView', () {
```
istzen, line 269: do we really need to test these methods for ride and drive?

fynsta: I also think that they should be moved to some trip_test.
 
But as a note: There you can only use a model of a specific type... So if I tested only for ride in that file, then I would not catch it if the behaviour were to be changed for drives in the future.

test/models/drive_test.dart
Showcas: Should I reduce the tests for equal? Or should we just leave them there since they already exist

fynsta: Let's leave them here, they don't hurt right now

test/models/trip_test.dart
```diff
@@ -0,0 +1,251 @@
...
+import '../util/mock_server.mocks.dart';
+
+void main() {
+  final MockUrlProcessor tripProcessor = MockUrlProcessor();
+  //setup muss in jeder Testklasse einmal aufgerufen werden
+  setUp(() async {
+    MockServer.setProcessor(tripProcessor);
+  });
+  group('Trip.duration', () {
+    test('returns the duration of a ride', () {
```
fynsta, line 18: okay but now can you write "trip" everywhere instead of ride/drive

Showcas: I renamed the variables. But I would leave the description cause they show that we cover rides and drives

fynsta: I meant in the group and test description

test/models/trip_test.dart
```diff
@@ -0,0 +1,229 @@
...
+import '../util/mock_server.mocks.dart';
+
+void main() {
+  final MockUrlProcessor tripProcessor = MockUrlProcessor();
+  //setup muss in jeder Testklasse einmal aufgerufen werden
+  setUp(() async {
+    MockServer.setProcessor(tripProcessor);
+  });
+  group('Trip.duration', () {
+    test('can handel a Ride', () {
```
fynsta, line 18: "handel" -> "handle"
"finsihed" -> "finished" (down below somewhere)

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+import 'package:motis_mitfahr_app/util/search/position.dart';
+
+import '../util/factories/drive_factory.dart';
+import '../util/factories/model_factory.dart';
+import '../util/factories/profile_factory.dart';
+import '../util/factories/ride_factory.dart';
+import '../util/mock_server.dart';
+import '../util/mock_server.mocks.dart';
+
+// Die Klasse UrlProcessor muss zu Beginn jeder Testdatei implementiert werden und die Methode processUrl überschrieben werden
```
Eddie-42, line 17: Please remove these comments

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+      expect(json['end_lng'], drive.endPosition.lng);
+      expect(json['end_time'], drive.endTime.toString());
+      expect(json['cancelled'], drive.cancelled);
+      expect(json['seats'], drive.seats);
+      expect(json['driver_id'], drive.driverId);
+      expect(json['hide_in_list_view'], drive.hideInListView);
+      expect(json.keys.length, 12);
+    });
+  });
+  group('Drive.approvedrides', () {
```
Eddie-42, line 148:
```suggestion
  group('Drive.approvedRides', () {
```

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+    });
+    test('return an empty List if rides has no approved rides', () async {
+      final Drive drive = DriveFactory().generateFake(
+        rides: [
+          RideFactory().generateFake(status: RideStatus.pending),
+        ],
+      );
+      expect(drive.approvedRides, []);
+    });
+    test('return a ride in a List, if rides only has 1 ride and it is approved', () async {
```
Eddie-42, line 163: Do we need this test? One approved ride is already the test below this, and I don't have reason to expect the outcome of this to be different

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+          RideFactory().generateFake(status: RideStatus.pending),
+          ride,
+          RideFactory().generateFake(status: RideStatus.rejected),
+          RideFactory().generateFake(status: RideStatus.cancelledByRider),
+          RideFactory().generateFake(status: RideStatus.withdrawnByRider),
+        ],
+      );
+      expect(drive.approvedRides, [ride]);
+    });
+    test('return correct rides with rides have more than 1 approved ones in it', () async {
```
Eddie-42, line 183: Please keep the test cases minimal. They should document the code and be readable. Every test case should be specific. Here, for example, there is no need for one Ride of every other status to be in the list, since it that has already been tested in the test above.

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+          RideFactory().generateFake(status: RideStatus.cancelledByDriver),
+          ride,
+          RideFactory().generateFake(status: RideStatus.rejected),
+          RideFactory().generateFake(status: RideStatus.cancelledByRider),
+          RideFactory().generateFake(status: RideStatus.withdrawnByRider),
+        ],
+      );
+      expect(drive.pendingRides, [ride]);
+    });
+    test('return correct rides with rides have more than 1 pending ride', () async {
```
Eddie-42, line 233: > Please keep the test cases minimal. They should document the code and be readable. Every test case should be specific. Here, for example, there is no need for one Ride of every other status to be in the list, since it that has already been tested in the test above.

Here, too

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+                )
+                .toJsonForApi()
+          ]));
+      expect(
+          await Drive.userHasDriveAtTimeRange(
+              DateTimeRange(start: DateTime.now(), end: DateTime.now().add(const Duration(hours: 10))), 2),
+          true);
+    });
+    test('returns false if there is no cancelled drive at the time range', () async {
+      when.call(driveProcessor.processUrl(any)).thenReturn(jsonEncode([
```
Eddie-42, line 271: This list can be shortened

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+      expect(drive.getMaxUsedSeats(), null);
+    });
+    test('returns 0 if there is no approved ride', () async {
+      final Drive drive = DriveFactory().generateFake(seats: 3, rides: <Ride>[
+        RideFactory().generateFake(
+            startTime: DateTime(2022, 2, 2, 14, 30),
+            endTime: DateTime(2022, 2, 2, 15),
+            status: RideStatus.pending,
+            seats: 1),
+        RideFactory().generateFake(
```
Eddie-42, line 331: Please remove this almost exact copy

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+      expect(drive.getMaxUsedSeats(), 0);
+    });
+    test('returns 1 if rides has only 1 approved ride with 1 seat', () async {
+      final Drive drive = DriveFactory().generateFake(seats: 3, rides: <Ride>[
+        RideFactory().generateFake(
+            startTime: DateTime(2022, 2, 2, 14, 30),
+            endTime: DateTime(2022, 2, 2, 15),
+            status: RideStatus.approved,
+            seats: 1),
+        RideFactory().generateFake(
```
Eddie-42, line 366: IMO, all of these further rides are unnecessary. We have established in the previous test that every ride with a status != approved will be counted as having 0 seats, so this is clutter

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+            startTime: DateTime(2022, 2, 2, 14, 30),
+            endTime: DateTime(2022, 2, 2, 15),
+            status: RideStatus.cancelledByDriver,
+            seats: 2),
+      ]);
+      expect(drive.getMaxUsedSeats(), 1);
+    });
+    test('returns 2 if rides has 2 approved ride with 1 seat at the same time', () async {
+      final Drive drive = DriveFactory().generateFake(seats: 3, rides: <Ride>[
+        RideFactory().generateFake(
```
Eddie-42, line 396: This, too. You want to establish that for rides at the same time, seats will be summed up. So I'd just create two rides at the same time, maybe one at a different time for good measure (even though that would arguably be a different test). Remove the rest, please. As a maintainer, this would be very time-consuming to read in order to understand the relatively basic message scenario this test conveys/tests

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+            seats: 1),
+        RideFactory().generateFake(
+            startTime: DateTime(2022, 2, 2, 14, 30),
+            endTime: DateTime(2022, 2, 2, 15),
+            status: RideStatus.pending,
+            seats: 1),
+      ]);
+      expect(drive.getMaxUsedSeats(), 2);
+    });
+    test('returns 4 if rides has a complex overlay where at its peak has 4 seats taken by approved rides', () async {
```
Eddie-42, line 429: I think in this case the point can be made that it's fair to test a more complex case, but please at least make a comment which rides overlap here so that I don't have to read all of this code (DateTimes aren't that easy to read)

Eddie-42: Also, and I think this goes for all of these tests, the description isn't really ideal, IMO. I would think about what I'd want to see in a console if a test fails to know exactly where it has failed, quickly. In this case, I'm not interested how many seats there are and what is expected; this information is already provided in the code once I go look for the test. What I'd want to know is that the complex Drive.getMaxUsedSeats test has failed. In the test above, I'd want to know that the overlapping getMaxUsedSeats test has failed, and in the test above that I'd want to know that the non-overlapping getMaxUsedSeats test has failed. I think naming tests like this might already help determining useful test cases

Eddie-42 (changes requested):
I think I've made my point regarding the tests using some examples. The same points could be made about some other tests, but I don't want to repeat myself too often. Please revise the tests from the point of view that it should be documenting the code and just as readable as the code. If anyone wants to discuss that with me, I'm willing to defend my position :) But 2000 lines is definitely too much for these tests IMO

test/models/drive_test.dart
```diff
@@ -0,0 +1,1003 @@
...
+      expect(drive.endTime, DateTime.parse(json['end_time']));
+      expect(drive.seats, json['seats']);
+      expect(drive.cancelled, json['cancelled']);
+      expect(drive.driverId, json['driver_id']);
+      expect(drive.driver, null);
+      expect(drive.rides, null);
+      expect(drive.hideInListView, json['hide_in_list_view']);
+    });
+    test('can handle associated models', () {
+      RideFactory().generateFakeJsonList(length: 3);
```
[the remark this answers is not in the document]
Showcas: I forgot about that one

Showcas:
> I think I've made my point regarding the tests using some examples. The same points could be made about some other tests, but I don't want to repeat myself too often. Please revise the tests from the point of view that it should be documenting the code and just as readable as the code. If anyone wants to discuss that with me, I'm willing to defend my position :) But 2000 lines is definitely too much for these tests IMO

I have redone the tests and tried to to take your points into account

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+import '../util/mock_server.mocks.dart';
+
+void main() {
+  final MockUrlProcessor tripProcessor = MockUrlProcessor();
+  //setup muss in jeder Testklasse einmal aufgerufen werden
+  setUp(() async {
+    MockServer.setProcessor(tripProcessor);
+  });
+
+  Object randomTripFactory() {
```
fynsta, line 22: the return type could be `TripFactory`, then you would not need the casting every time

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+
+import '../util/factories/drive_factory.dart';
+import '../util/factories/ride_factory.dart';
+import '../util/factories/trip_factory.dart';
+import '../util/mock_server.dart';
+import '../util/mock_server.mocks.dart';
+
+void main() {
+  final MockUrlProcessor tripProcessor = MockUrlProcessor();
+  //setup muss in jeder Testklasse einmal aufgerufen werden
```
fynsta, line 17: I would remove this comment everywhere, it's not even true

test/util/factories/trip_factory.dart
```diff
@@ -16,5 +16,6 @@ abstract class TripFactory<T extends Trip> extends ModelFactory<T> {
     DateTime? endTime,
     int? seats,
     bool createDependencies = true,
+    bool hideInListView = false,
```
fynsta, line 19: can you move this to the spot where it is in the class definition? Definitely above `createDependencies`, that should always be the last argument for clarity

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+      final Trip trip = (randomTripFactory() as TripFactory).generateFake(
+          startTime: now,
+          endTime: now.add(
+            const Duration(hours: 2),
+          ));
+      expect(trip.duration, const Duration(hours: 2));
+    });
+  });
+  group('Trip.isFinished', () {
+    test('trip is finsihed', () {
```
Eddie-42, line 40:
```suggestion
    test('trip is finished', () {
```

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+    });
+    test('not showing in ListView', () {
+      final Trip trip = (randomTripFactory() as TripFactory).generateFake(
+        startTime: DateTime.now().subtract(const Duration(hours: 4)),
+        endTime: DateTime.now().subtract(const Duration(hours: 2)),
+      );
+      expect(trip.shouldShowInListView(past: false), false);
+    });
+  });
+  group('Trip.equals', () {
```
Eddie-42, line 204: Am I wrong or does this not have a test case that is actually equal? :D

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+      final Trip drive = DriveFactory().generateFake(
+        cancelled: true,
+        startTime: DateTime.now().subtract(const Duration(hours: 2)),
+        endTime: DateTime.now().add(const Duration(hours: 4)),
+      );
+      expect(drive.shouldShowInListView(past: true), false);
+    });
+  });
+  group('Trip.shouldShowInListView', () {
+    test('shows in ListView', () {
```
Eddie-42, line 174: This kind of sounds like a widget test. I would try to go for something more descriptive of the test case, like "future trip in future list view" or something

Eddie-42: Also, what is the difference between this and the `Trip.shouldShowInListView` group before?

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+      );
+      expect(ride.shouldShowInListView(past: false), false);
+      final Trip drive = DriveFactory().generateFake(
+        hideInListView: true,
+        startTime: DateTime.now().add(const Duration(hours: 2)),
+        endTime: DateTime.now().add(const Duration(hours: 4)),
+      );
+      expect(drive.shouldShowInListView(past: false), false);
+    });
+    test('returns true for past finsihed trip', () {
```
Eddie-42, line 144:
```suggestion
    test('returns true for past finished trip', () {
```

Please use a code spell checker :) I'm not going to comment on spelling mistakes anymore from here on out

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+          startTime: DateTime.now().subtract(const Duration(hours: 6)),
+          endTime: DateTime.now().subtract(const Duration(hours: 2)));
+      final Trip trip2 = (randomTripFactory() as TripFactory).generateFake(
+          startTime: DateTime.now().subtract(const Duration(hours: 3)),
+          endTime: DateTime.now().add(const Duration(hours: 2)));
+      expect(trip1.overlapsWith(trip2), true);
+    });
+  });
+  group('Trip.overlapswithRange', () {
+    test('trip is after range', () {
```
Eddie-42, line 94: What's with trip is before range?

test/models/trip_test.dart
```diff
@@ -0,0 +1,369 @@
...
+    });
+    test('trip is upcoming', () {
+      final Trip trip = (randomTripFactory() as TripFactory).generateFake(
+          startTime: DateTime.now().add(const Duration(hours: 2)),
+          endTime: DateTime.now().add(const Duration(hours: 6)));
+      expect(trip.isOngoing, false);
+    });
+  });
+  group('Trip.overlapsWith', () {
+    test('trip is before other', () {
```
Eddie-42, line 74: What's with trip is after other?

test/models/drive_test.dart
```diff
@@ -0,0 +1,489 @@
...
+    test('can handle null', () async {
+      final Drive drive = DriveFactory().generateFake(
+        createDependencies: false,
+      );
+      expect(drive.approvedRides, null);
+    });
+    test('has no approved rides', () async {
+      final Drive drive = DriveFactory().generateFake(
+        rides: [
+          RideFactory().generateFake(status: RideStatus.pending),
```
Eddie-42, line 151: IMO, what you could very well do here is have this list contain rides from every status EXCEPT approved, to prove the point that only approved rides will be counted. I think that could even be a neat one-liner, like 
```
rides: List.generate(
            RideStatus.values.length, (index) => RideFactory().generateFake(status: RideStatus.values[index]))
            .where((element) => element.status != RideStatus.approved)
            .toList(),
```

test/models/drive_test.dart
```diff
@@ -0,0 +1,489 @@
...
+        ],
+      );
+      expect(drive.approvedRides, []);
+    });
+    test('has approved rides', () async {
+      final Ride ride1 = RideFactory().generateFake(status: RideStatus.approved);
+      final Ride ride2 = RideFactory().generateFake(status: RideStatus.approved);
+
+      final Drive drive = DriveFactory().generateFake(
+        rides: [
```
Eddie-42, line 162: Same here and in `Drive.pendingRides`

test/models/drive_test.dart
```diff
@@ -0,0 +1,489 @@
...
+  });
+  group('Drive.getMaxUsedSeats', () {
+    test('can handle null', () async {
+      final Drive drive = DriveFactory().generateFake(createDependencies: false);
+      expect(drive.getMaxUsedSeats(), null);
+    });
+    test('no seats taken', () async {
+      final Drive drive = DriveFactory().generateFake(seats: 3, rides: <Ride>[
+        RideFactory().generateFake(
+            startTime: DateTime(2022, 2, 2, 14, 30),
```
Eddie-42, line 276: Could you please continue using `DateTime.now()` here and everywhere else? I think this is really illegible

test/models/drive_test.dart
```diff
@@ -0,0 +1,489 @@
...
+            startTime: DateTime(2022, 2, 2, 14, 20),
+            endTime: DateTime(2022, 2, 2, 15),
+            status: RideStatus.approved,
+            seats: 2),
+      ]);
+      expect(drive.getMaxUsedSeats(), 3);
+    });
+  });
+  group('Drive.isRidePossible', () {
+    test('can handle empty drive', () async {
```
Eddie-42, line 340: Please remove `can handle` from these descriptions. You could write these before any test case and I don't think it adds any value (and in this case, it's inconsistent with the description right below)

test/models/drive_test.dart
```diff
@@ -0,0 +1,489 @@
...
+      end: 'Frankfurt',
+      endPosition: Position(2, 2),
+      endTime: DateTime(2022, 12),
+      seats: 1,
+      cancelled: false,
+      driverId: 2,
+      createDependencies: false,
+    );
+    test('same drive', () async {
+      expect(drive0.equals(drive0), true);
```
Eddie-42, line 448: I think you should use a deepcopy here. That would show that we're using value equality, not reference equality. For Ride too

test/models/drive_test.dart
```diff
@@ -0,0 +1,489 @@
...
+    });
+    test('handle ride', () async {
+      final Ride ride = RideFactory().generateFake(createDependencies: false);
+      expect(drive0.equals(ride), false);
+    });
+    test('different cancelled', () async {
```
Eddie-42, line 454: I don't like this test. Either we test this through for every parameter (which I wouldn't do), or we just do it for one, and I think that's preferably the ID. The case can be made that it is important to show that even drives with the same ID are different to us if they have different values, but then why cancelled? I think start or something less obscure would make more sense in that case (you already have driverId below, that's fine too)

test/models/ride_test.dart
```diff
@@ -0,0 +1,468 @@
...
+      expect(json['status'], ride.status.index);
+      expect(json['seats'], ride.seats);
+      expect(json['price'], ride.price);
+      expect(json['drive_id'], ride.driveId);
+      expect(json['rider_id'], ride.riderId);
+      expect(json['hide_in_list_view'], ride.hideInListView);
+      expect(json.keys.length, 14);
+    });
+  });
+  group('Ride.toJsonforApi', () {
```
Eddie-42, line 178: Do we test this now? I think tests for testing infrastructure are a bit much, we're already getting the coverage by calling this 100 times in tests anyway.

test/models/ride_test.dart
```diff
@@ -0,0 +1,468 @@
...
+              DateTimeRange(
+                start: DateTime.now().add(const Duration(hours: 6)),
+                end: DateTime.now().add(const Duration(hours: 8)),
+              ),
+              2),
+          true);
+    });
+  });
+  group('Ride.shouldShowInListView', () {
+    test('show in Listview', () {
```
Eddie-42, line 287:
```suggestion
    test('show in ListView', () {
```
Also, isn't this already covered by the trip test?

test/models/ride_test.dart
```diff
@@ -0,0 +1,468 @@
...
+      price: NullableParameter(6),
+    );
+    test('same ride', () async {
+      expect(ride0.equals(ride0), true);
+    });
+    test('handle drive', () async {
+      final Drive drive = DriveFactory().generateFake(createDependencies: false);
+      expect(ride0.equals(drive), false);
+    });
+    test('different status', () async {
```
Eddie-42, line 333: Here you're testing this for a few random parameters again. I would suggest testing it for the ID and maybe one additional parameter, (or, if you want to test every parameter, do that and do it in a loop, please)

Eddie-42 (commented):
I think I am very harsh in this review, but I do think this is a great improvement over the previous iteration and after this is all fixed, I am happy to approve. Very good job!

test/models/trip_test.dart
Showcas: It does not see RideFactory or RideFactory as an instance of TripFactory when I declare TripFactory as a Return type. The reason could be since they don't have a constructor.

test/models/trip_test.dart
Showcas: missed that one, removed it in all the others

test/models/trip_test.dart
Showcas: Forgot to add that one here

test/models/trip_test.dart
Showcas: forgot to remove the old one

test/models/drive_test.dart
Showcas: that is covered by the equals test in trip. since euquals in drive and ride call their super from trip. I only check for the extra checks in drive and ride and the core trip values in Trip.equals

test/models/ride_test.dart
Showcas: I can take it out if you want. But I don't think it hurts

test/models/ride_test.dart
Showcas: here are new cases since it is overwriten in ride. In drive it is not overwriten so it is not tested in drive

test/util/factories/mock_trip_factory.dart
```diff
@@ -0,0 +1,49 @@
+import 'package:motis_mitfahr_app/util/search/position.dart';
+import 'package:motis_mitfahr_app/util/trip/trip.dart';
+
+import 'trip_factory.dart';
+
+class MockTrip extends Trip {
```
fynsta, line 6: I'm not feeling great about two things: 

- this file/factory does not fit with the others
- the fact that we need to repeat all fields here _again_... 

Maybe the best solution is just to remove the `abstract` keyword from Trip, and then use the `TripFactory` for our purposes. I think that would solve a lot of complication, and `abstract` is currently really unnecessary since `Trip` implements all its methods.

What do you think @Eddie-42 ?

fynsta: I have pushed my idea on another branch. Please see this commit: https://github.com/MOTIS-Mitfahr-App/flutter-app/commit/283025be1f69fc9670e73eb2b3ae855b10b15adc

I think it's a lot better?

Showcas: I agree

test/models/trip_test.dart
```diff
@@ -0,0 +1,203 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:motis_mitfahr_app/util/search/position.dart';
+import 'package:motis_mitfahr_app/util/trip/trip.dart';
+
+import '../util/factories/mock_trip_factory.dart';
+import '../util/mock_server.dart';
+import '../util/mock_server.mocks.dart';
+
+void main() {
+  final MockUrlProcessor tripProcessor = MockUrlProcessor();
```
istzen, line 11: can you rebase? then this needs to be a MockRequestProcessor if i am correct

fynsta: as written above the rebase is broken 😢  Maybe I'll have to open a new branch and copy the changes by hand.

test/models/drive_test.dart
Eddie-42: Mh okay I suspect we'll have to unify that anyway some time

test/models/drive_test.dart
Eddie-42: Think you forgot that one because GitHub kind of hides it ;)

Eddie-42 (approved):

test/models/trip_test.dart
istzen: ah sorry my mistake. You could do that i guess or is there any other way?

istzen: maybe that's the best way idk if there is another one

fynsta (approved):

MarkFalke (approved):

fynsta:
Superseded by #97

PR #87, opened 2023-01-23
Improve request mocking

fynsta:
This overhauls the way that the request mocking works. Previously we could only match the request url. Now I also added the possibility to mock and verify the request method and body. This should be very helpful to write better assertions, because there's a way to differentiate between "GET" and "PATCH" calls, and to assert the body of that PATCH call.

For this, there are four new methods that should be used (what do you think about those names?):

- `whenRequest` instead of `when`
- `thenReturnJson` instead of `thenReturn` (if you want to return json)
- `verifyRequest` instead of `verify`
- `verifyRequestNever` instead of `verifyNever`

The syntax of how exactly to use them should hopefully clear from the changes I made in the test pages. The new method will also do the json encoding for us, because that was a bit annoying before.


test/util/request_processor.dart
```diff
@@ -0,0 +1,68 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/annotations.dart';
+import 'package:mockito/mockito.dart';
+
+import 'request_processor.mocks.dart';
+
+@GenerateMocks([RequestProcessor])
+class RequestProcessor {
+  String process(String url, {String? method, Map<String, dynamic>? body}) {
+    throw UnimplementedError();
```
fynsta, line 12: this now throws an error, just to mak sure that we know when we forgot to mock one of the HTTP calls.

test/models/message_test.dart
```diff
@@ -21,10 +23,18 @@ void main() {
       final message = MessageFactory().generateFake(
         read: false,
       );
+      whenRequest(messageProcessor, '');
+
       await message.markAsRead();
-      SupabaseManager.supabaseClient.from('messages').update({
```
fynsta, line 25: @istzen I stumbled over this when rewriting the mocking. This line should be unnecessary. There is no point in calling supabase, because HTTP is disabled and the only response will be our own mocks.

If you need to "change data" in supabase, you need to change the object and then renew the mock that returns the object on a GET request (see my RideDetail tests for example)

But in this case, that's not needed either. "message.read = true" is already set by the method itself. And we only need to know that the rpc function was called. You can see that verification below (note that it needed to be mocked in line 26 before the method call). We can't really test whether the supabase function actually marks the message as read, but that's not our job.

istzen: yes thanks, i think that was the work of my friend and helper Copilot.

test/example_test.dart
```diff
@@ -1,28 +1,26 @@
-import 'dart:convert';
```
istzen, line 1: do we still need this file?

istzen (commented):
i very much like the new methods, they will help a lot. The names are fine i think. Maybe we could add some comments so it's clear whats happening but not sure if thats really nessecary

fynsta:
Fyi I just noticed that `whenRequest` should probably have the same return type as `when`, otherwise that would be confusing. So I added `thenReturnJson` to help with the json encoding instead.

test/example_test.dart
fynsta: No, but I think @Showcas had removed it in his PR. So I did not want to create any more conflicts between the files.

istzen: yes a reviewed that after yours my mistkae

istzen (approved):

MarkFalke (approved):

Showcas (approved):

Eddie-42 (approved):

PR #89, opened 2023-01-24
Rides page/drives page tests

istzen:
add test for drivesPage and RidesPage, most of them i refactored into trip_page_builder_test since the code is the same that should not be a Problem.

### DIFF
--- a/test/widgets/drives_page_test.dart
+++ b/test/widgets/drives_page_test.dart
@@ -22,7 +22,7 @@ void main() {
     MockServer.setProcessor(processor);
   });
   setUp(() async {
-    profile = ProfileFactory().generateFake();
+    profile = ProfileFactory().generateFake(id: 1);
     SupabaseManager.setCurrentProfile(profile);
     whenRequest(processor).thenReturn(jsonEncode([]));
   });
@@ -31,6 +31,10 @@ void main() {
     final List<RealtimeChannel> subscription = SupabaseManager.supabaseClient.getChannels();
     expect(subscription.length, 1);
     expect(subscription[0].topic, 'realtime:public:drives:1');
+    verifyRequest(
+      processor,
+      urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.1&order=start_time.asc.nullslast'),
+    ).called(1);
   });
 
   testWidgets('FAB works', (WidgetTester tester) async {
--- a/test/widgets/drives_page_test.dart
+++ b/test/widgets/drives_page_test.dart
@@ -1,0 +1,47 @@
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:motis_mitfahr_app/account/models/profile.dart';
+import 'package:motis_mitfahr_app/drives/pages/create_drive_page.dart';
+import 'package:motis_mitfahr_app/drives/pages/drives_page.dart';
+import 'package:motis_mitfahr_app/util/supabase.dart';
+import 'package:supabase_flutter/supabase_flutter.dart';
+
+import '../util/factories/profile_factory.dart';
+import '../util/mock_server.dart';
+import '../util/pump_material.dart';
+import '../util/request_processor.dart';
+import '../util/request_processor.mocks.dart';
+
+void main() {
+  late Profile profile;
+  final MockRequestProcessor processor = MockRequestProcessor();
+
+  setUpAll(() async {
+    MockServer.setProcessor(processor);
+  });
+  setUp(() async {
+    profile = ProfileFactory().generateFake();
+    SupabaseManager.setCurrentProfile(profile);
+    whenRequest(processor).thenReturn(jsonEncode([]));
+  });
+  testWidgets('has stream subscription', (WidgetTester tester) async {
+    await pumpMaterial(tester, const DrivesPage());
+    final List<RealtimeChannel> subscription = SupabaseManager.supabaseClient.getChannels();
+    expect(subscription.length, 1);
+    expect(subscription[0].topic, 'realtime:public:drives:1');
+  });
+
+  testWidgets('FAB works', (WidgetTester tester) async {
+    await pumpMaterial(tester, const DrivesPage());
+    expect(find.byType(CircularProgressIndicator), findsOneWidget);
+    await tester.pump();
+    expect(find.byType(CircularProgressIndicator), findsNothing);
+    final Finder fab = find.byKey(const Key('drivesFAB'));
+    expect(fab, findsOneWidget);
+    await tester.tap(fab);
+    await tester.pumpAndSettle();
+    expect(find.byType(CreateDrivePage), findsOneWidget);
+  });
+}
--- a/test/widgets/drives_page_test.dart
+++ b/test/widgets/drives_page_test.dart
@@ -33,7 +33,7 @@ void main() {
     expect(subscription[0].topic, 'realtime:public:drives:1');
     verifyRequest(
       processor,
-      urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.1&order=start_time.asc.nullslast'),
+      urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.${profile.id}&order=start_time.asc.nullslast'),
     ).called(1);
   });
 
--- a/test/widgets/trip_page_builder_test.dart
+++ b/test/widgets/trip_page_builder_test.dart
@@ -90,7 +90,8 @@ void main() {
     });
     testWidgets('shows upcoming Trips at beginning', (WidgetTester tester) async {
       whenRequest(processor,
-              urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.1&order=start_time.asc.nullslast'))
+              urlMatcher:
+                  equals('/rest/v1/drives?select=%2A&driver_id=eq.${profile.id}&order=start_time.asc.nullslast'))
           .thenReturnJson([]);
       await pumpMaterial(tester, const DrivesPage());
       expect(find.byKey(const Key('upcomingTrips')), findsOneWidget);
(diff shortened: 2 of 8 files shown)


test/widgets/drives_page_test.dart
```diff
@@ -0,0 +1,47 @@
...
+void main() {
+  late Profile profile;
+  final MockRequestProcessor processor = MockRequestProcessor();
+
+  setUpAll(() async {
+    MockServer.setProcessor(processor);
+  });
+  setUp(() async {
+    profile = ProfileFactory().generateFake();
+    SupabaseManager.setCurrentProfile(profile);
```
fynsta, line 26: This is unnecessary right? You are not using the profile

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,270 @@
...
+            drive: NullableParameter(null),
+            driveId: 4,
+            createDependencies: false,
+          )
+          .toJsonForApi(),
+    ];
+    whenRequest(processor, urlMatcher: equals('/rest/v1/rides?select=%2A&rider_id=eq.1&order=start_time.asc.nullslast'))
+        .thenReturnJson(rides);
+    whenRequest(processor, urlMatcher: matches(RegExp(r'/rest/v1/drives.*id=eq\.3')), methodMatcher: equals('GET'))
+        .thenReturnJson(DriveFactory().generateFake(id: 1).toJsonForApi());
```
fynsta, line 65: The id here should probably match up with the prevopis driveId to be a bit more realistic. You can create the drive before the call and then use it throughout the test

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,270 @@
...
+    testWidgets('finds both Tabs(before and after Drives are loaded)', (WidgetTester tester) async {
+      whenRequest(processor,
+              urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.1&order=start_time.asc.nullslast'))
+          .thenReturnJson([]);
+      await pumpMaterial(tester, const DrivesPage());
+      final Finder tabBar = find.byType(TabBar);
+      expect(tabBar, findsOneWidget);
+      expect(find.descendant(of: tabBar, matching: find.byType(Tab)), findsNWidgets(2));
+      await tester.pump();
+      final Finder tabBarLoaded = find.byType(TabBar);
```
fynsta, line 86: you don't need to redefine the finder, just write another expectation

test/widgets/drives_page_test.dart
istzen: No i Need it, since i'm getting all rides/drives for current user. I could replace the id however With the supabase.currentProfile.id of you want

test/widgets/trip_page_builder_test.dart
istzen: yes thanks. forgot to do that.

test/widgets/trip_page_builder_test.dart
istzen: yes okay wasn't sure

test/widgets/trip_page_builder_test.dart
istzen: yes thanks. forgot to do that.

test/widgets/drives_page_test.dart
fynsta: Where? This whole test file does not use the profile. For the TripPageBuilder, I would agree.

istzen: yes but when the cards get build the current Profile is accessed and it's easier if i know the id

istzen: i gave the profile a purpose.

test/widgets/trip_page_builder_test.dart
istzen: now that i'm thinking why is this relevant? that has nothing to do the the Widgets I'm testing, but with the cards.

fynsta: Well we should return consistent data via the api? Who knows what our code relies on / will rely on.

istzen: but here I'm returning drives for my ride and in the other cases drives for myself so shouldn't they be different

test/widgets/drives_page_test.dart
```diff
@@ -0,0 +1,51 @@
...
+    whenRequest(processor).thenReturn(jsonEncode([]));
+  });
+  testWidgets('has stream subscription', (WidgetTester tester) async {
+    await pumpMaterial(tester, const DrivesPage());
+    final List<RealtimeChannel> subscription = SupabaseManager.supabaseClient.getChannels();
+    expect(subscription.length, 1);
+    expect(subscription[0].topic, 'realtime:public:drives:1');
+    verifyRequest(
+      processor,
+      urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.1&order=start_time.asc.nullslast'),
```
fynsta, line 36: you wanted me to comment that. I would be more comfortable if you checked for
```
/rest/v1/drives?select=%2A&driver_id=eq.${profile.id}&order=start_time.asc.nullslast
```
Otherwise you are always confused about where the random "1" is coming from. And maybe there will be an off-by-one error or something that we would catch with that.

There are some other places where you should probably also add a non-hardcoded id.

test/widgets/trip_page_builder_test.dart
fynsta: Maybe I should be more clear about what I mean: You are explicitly loading "the drive with id=3" in the request. Then you return a drive with id=1. That's just weird.

istzen: ah yes sorry my mistake

test/widgets/drives_page_test.dart
istzen: yes did that

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,278 @@
...
+      final Finder tabBar = find.byType(TabBar);
+      expect(tabBar, findsOneWidget);
+      expect(find.descendant(of: tabBar, matching: find.byType(Tab)), findsNWidgets(2));
+      await tester.pump();
+      expect(tabBar, findsOneWidget);
+      expect(find.descendant(of: tabBar, matching: find.byType(Tab)), findsNWidgets(2));
+    });
+    testWidgets('shows upcoming Trips at beginning', (WidgetTester tester) async {
+      whenRequest(processor,
+              urlMatcher: equals('/rest/v1/drives?select=%2A&driver_id=eq.1&order=start_time.asc.nullslast'))
```
fynsta, line 93: missed profile id here

istzen: thanks

fynsta (approved):

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,279 @@
...
+            hideInListView: false,
+            rider: NullableParameter(null),
+            drive: NullableParameter(null),
+            driveId: 7,
+            createDependencies: false,
+          )
+          .toJsonForApi(),
+    ];
+    whenRequest(processor,
+            urlMatcher: equals('/rest/v1/rides?select=%2A&rider_id=eq.${profile.id}&order=start_time.asc.nullslast'))
```
fynsta, line 63: trailing comma

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,279 @@
...
+    profile = ProfileFactory().generateFake(id: 1);
+    SupabaseManager.setCurrentProfile(profile);
+  });
+
+  // since the code for the TripPageBuilder is the same for both pages,
+  // we only test the DrivesPage and assume that the RidesPage works the same.
+  // This test is to show that if the ridesPage is called there are rideCards shown.
+  testWidgets('shows Ride card for RidesPage', (WidgetTester tester) async {
+    final List<Map<String, dynamic>> rides = [
+      RideFactory()
```
fynsta, line 39: I still think this should be saved in a variable like `ride1`, and then the url later should use `eq.id=${ride1.driveId}`. Otherwise you would have to change both places if you wanted something different. And the `id` and `driveId` could be autogenerated which would cover more cases

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,279 @@
...
+          .generateFake(
+            id: 1,
+            endTime: DateTime.now().add(const Duration(days: 1)),
+            hideInListView: false,
```
fynsta, line 43: I mean, that's the default? Why even set it?

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,279 @@
...
+        .thenReturnJson(DriveFactory().generateFake(id: 7).toJsonForApi());
+
+    await pumpMaterial(tester, const RidesPage());
+    await tester.pump();
+    final Finder rideCard = find.byType(RideCard);
+    expect(rideCard, findsNWidgets(2));
+  });
+
+  group('tabs', () {
+    setUp(() async {});
```
fynsta, line 77: can be removed

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,279 @@
...
+    testWidgets('finds both Tabs(before and after Drives are loaded)', (WidgetTester tester) async {
+      whenRequest(processor,
+              urlMatcher:
```
fynsta, line 80: also trailing comma

test/widgets/trip_page_builder_test.dart
```diff
@@ -0,0 +1,279 @@
...
+                  equals('/rest/v1/drives?select=%2A&driver_id=eq.${profile.id}&order=start_time.asc.nullslast'))
+          .thenReturnJson([]);
+      await pumpMaterial(tester, const DrivesPage());
+      await tester.pump();
+      expect(find.byType(DriveCard), findsNothing);
+      expect(find.byKey(const Key('emptyMessage')), findsOneWidget);
+    });
+  });
+
+  group('shows Cards past', () {
```
fynsta, line 195: let's call the group "Past Cards" or something, I think only the test cases themselves should contain verbs

test/widgets/trip_page_builder_test.dart
istzen: to make it better to read.

Showcas (approved):

Eddie-42 (approved):

PR #90, opened 2023-01-24
Home page

MarkFalke:
Finally the Homepage is here.
The Homepage has a box in which the user gets notifications if anything related to him has updated.
Below are two buttons, making the create drive and search ride sequence easier.

### DIFF
--- a/lib/home_page.dart
+++ b/lib/home_page.dart
@@ -35,18 +35,25 @@ class _HomePageState extends State<HomePage> {
     SupabaseManager.supabaseClient.channel('public:messages').on(
       RealtimeListenTypes.postgresChanges,
       ChannelFilter(event: 'INSERT', schema: 'public', table: 'messages', filter: 'sender_id=neq.$profileId'),
+      // ignore: always_specify_types
       (payload, [ref]) {
+        // ignore: avoid_dynamic_calls
         if (payload['new']['sender_id'] != profileId) {
+          // ignore: avoid_dynamic_calls
           _loadNewMessage(payload['new']['id']);
         }
       },
     ).on(
       RealtimeListenTypes.postgresChanges,
       ChannelFilter(event: 'UPDATE', schema: 'public', table: 'messages', filter: 'sender_id=neq.$profileId'),
+
+      // ignore: always_specify_types
       (payload, [ref]) {
+        // ignore: avoid_dynamic_calls
         if (payload['new']['read'] == true) {
           setState(() {
             _items =
+                // ignore: avoid_dynamic_calls
                 _items.where((Model element) => element is RideEvent || element.id != payload['new']['id']).toList();
           });
         }
@@ -55,17 +62,23 @@ class _HomePageState extends State<HomePage> {
     SupabaseManager.supabaseClient.channel('public:ride_events').on(
       RealtimeListenTypes.postgresChanges,
       ChannelFilter(event: 'INSERT', schema: 'public', table: 'ride_events'), //, filter: 'sender_id=neq.$profileId'
+      // ignore: always_specify_types
       (payload, [ref]) {
+        // ignore: avoid_dynamic_calls
         if (payload['new']['sender_id'] != profileId) {
+          // ignore: avoid_dynamic_calls
           _loadNewRideEvent(payload['new']['id']);
         }
       },
     ).on(
       RealtimeListenTypes.postgresChanges,
       ChannelFilter(event: 'UPDATE', schema: 'public', table: 'ride_events'), //, filter: 'sender_id=neq.$profileId'
+      // ignore: always_specify_types
       (payload, [ref]) {
+        // ignore: avoid_dynamic_calls
         if (payload['new']['read'] == true) {
           setState(() {
+            // ignore: avoid_dynamic_calls
             _items = _items.where((Model element) => element is Message || element.id != payload['new']['id']).toList();
           });
         }
--- a/lib/home_page.dart
+++ b/lib/home_page.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 
 import 'util/buttons/button.dart';
 import 'util/chat/models/message.dart';
@@ -15,6 +16,7 @@ class HomePage extends StatefulWidget {
 class _HomePageState extends State<HomePage> {
   late Stream<List<RideEvent>> _events;
   late Stream<List<Message>> _messages;
+  bool _fullyLoaded = false;
 
   @override
   void initState() {
@@ -33,21 +35,30 @@ class _HomePageState extends State<HomePage> {
         .order('created_at')
         .map(
           (List<Map<String, dynamic>> rideEvents) => RideEvent.fromJsonList(
-              rideEvents.where((Map<String, dynamic> rideEvent) => rideEvent['sender_id'] != profileId).toList()),
+            rideEvents.where((Map<String, dynamic> rideEvent) => rideEvent['sender_id'] != profileId).toList(),
+          ),
         );
-    _messages =
-        SupabaseManager.supabaseClient.from('messages').stream(primaryKey: <String>['id']).eq('read', false).map(
-              (List<Map<String, dynamic>> messages) => Message.fromJsonList(
-                  messages.where((Map<String, dynamic> rideEvent) => rideEvent['sender_id'] != profileId).toList()),
-            );
+    _messages = SupabaseManager.supabaseClient
+        .from('messages')
+        .stream(primaryKey: <String>['id'])
+        .eq('read', false)
+        .order('created_at')
+        .map(
+          (List<Map<String, dynamic>> messages) => Message.fromJsonList(
+            messages.where((Map<String, dynamic> message) => message['sender_id'] != profileId).toList(),
+          ),
+        );
+    setState(() {
+      _fullyLoaded = true;
+    });
   }
 
   @override
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-          // title: Text(S.of(context).homePageTitle),
-          ),
+        title: Text(S.of(context).pageHomeTitle),
+      ),
       body: Center(
         child: Column(
           children: <Widget>[
@@ -90,7 +101,7 @@ class _HomePageState extends State<HomePage> {
                   return const SizedBox(height: 12);
                 },
                 itemBuilder: (BuildContext context, int index) {
-                  return buildcard(index, context);
+                  return;
                 },
               ),
             )
@@ -107,17 +118,17 @@ class _HomePageState extends State<HomePage> {
   }
 }
 
-Widget buildcard(int index, context) => Container(
-      color: Colors.blueAccent,
-      width: double.infinity,
-      height: 50,
-      child: const Center(
-        child: InkWell(
-          onTap: () => Navigator.of(context).push(
-            MaterialPageRoute(
-              builder: (BuildContext context) => DriveDetailPage.fromDrive(trip),
-            ),
-          ),
-        ),
-      ),
-    );
+// Widget buildcard(int index, context) => Container(
+//       color: Colors.blueAccent,
+//       width: double.infinity,
+//       height: 50,
+//       child: const Center(
+//         child: InkWell(
+//           onTap: () => Navigator.of(context).push(
+//             MaterialPageRoute(
+//               builder: (BuildContext context) => DriveDetailPage(trip),
+//             ),
+//           ),
+//         ),
+//       ),
+//     );
--- a/lib/home_page.dart
+++ b/lib/home_page.dart
@@ -1,5 +1,10 @@
 import 'package:flutter/material.dart';
 
+import 'util/buttons/button.dart';
+import 'util/chat/models/message.dart';
+import 'util/ride_event.dart';
+import 'util/supabase.dart';
+
 class HomePage extends StatefulWidget {
   const HomePage({super.key});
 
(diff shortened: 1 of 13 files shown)


fynsta:
Do you see the Problems Tab in your VSCode? It should show ~20 problems.

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+        }
+      },
+    ).on(
+      RealtimeListenTypes.postgresChanges,
+      ChannelFilter(event: 'UPDATE', schema: 'public', table: 'messages', filter: 'sender_id=neq.$profileId'),
+      (payload, [ref]) {
+        if (payload['new']['read'] == true) {
+          setState(() {
+            _items =
+                _items.where((Model element) => element is RideEvent || element.id != payload['new']['id']).toList();
```
fynsta, line 50: I would prefer `element is! Message`, that should be better for the future, in case we add other types of info

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+          });
+        }
+      },
+    ).subscribe();
+    SupabaseManager.supabaseClient.channel('public:ride_events').on(
+      RealtimeListenTypes.postgresChanges,
+      ChannelFilter(event: 'INSERT', schema: 'public', table: 'ride_events'), //, filter: 'sender_id=neq.$profileId'
```
fynsta, line 57: can these comments be removed?

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+          _loadNewRideEvent(payload['new']['id']);
+        }
+      },
+    ).on(
+      RealtimeListenTypes.postgresChanges,
+      ChannelFilter(event: 'UPDATE', schema: 'public', table: 'ride_events'), //, filter: 'sender_id=neq.$profileId'
+      (payload, [ref]) {
+        if (payload['new']['read'] == true) {
+          setState(() {
+            _items = _items.where((Model element) => element is Message || element.id != payload['new']['id']).toList();
```
fynsta, line 82: I think you might be looking for  `removeWhere`: https://api.flutter.dev/flutter/dart-core/List/removeWhere.html

lib/home_page.dart
```diff
@@ -35,18 +35,25 @@ class _HomePageState extends State<HomePage> {
     SupabaseManager.supabaseClient.channel('public:messages').on(
       RealtimeListenTypes.postgresChanges,
       ChannelFilter(event: 'INSERT', schema: 'public', table: 'messages', filter: 'sender_id=neq.$profileId'),
+      // ignore: always_specify_types
```
fynsta, line 38: By resolving, I don't mean "ignoring". I'm sure these type ones are especially easily fixable. And for nested Json down below, we already have the `deep_pick`package: https://pub.dev/packages/deep_pick

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+  }
+
+  Future<void> _loadNewMessage(int messageId) async {
+    final Map<String, dynamic> data = await SupabaseManager.supabaseClient.from('messages').select('''
+      *,
+      sender: sender_id(*)
+      )
+    ''').eq('id', messageId).single();
+    final Message message = Message.fromJson(data);
+    if (message.read == false) {
```
fynsta, line 243: if(!message.read)

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+  Widget _buildRideEventWidget(RideEvent rideEvent, BuildContext context) {
+    final bool isforDrive = rideEvent.ride!.riderId != SupabaseManager.getCurrentProfile()!.id!;
+    return Card(
+      child: InkWell(
+        child: ListTile(
+          leading: isforDrive ? const Icon(Icons.drive_eta) : const Icon(Icons.chair),
+          title: Text(rideEvent.getTitle(context)),
+          subtitle: Text(rideEvent.getMessage(context)),
+          trailing: Text(
+            DateFormat('HH:mm').format(rideEvent.createdAt!),
```
fynsta, line 285: don't we have a time formatter in the `localeManager`?

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+              ),
+            );
+          },
+        ),
+      ),
+    );
+  }
+
+  Widget _buildRideEventWidget(RideEvent rideEvent, BuildContext context) {
+    final bool isforDrive = rideEvent.ride!.riderId != SupabaseManager.getCurrentProfile()!.id!;
```
fynsta, line 277: `rideEvent.ride!.rider.isCurrentUser`

lib/l10n/app_de.arb
```diff
@@ -330,6 +330,54 @@
...
   "pageChatMessageBarHint": "Schreibe eine Nachricht",
 
+  "pageHomeTitle": "Home",
+  "pageHomeSearchButton": "Mitfahrt suchen",
+  "pageHomeCreateButton": "Fahrt erstellen",
+
+  "rideEventPendingTitle": "Mitfahrt angefragt!",
+  "rideEventApprovedTitle": "Mitfahrt bestätigt!",
+  "rideEventRejectedTitle": "Mitfahrt abgelehnt!",
+  "rideEventWithdrawnTitle": "Mitfahrt verworfen!",
```
fynsta, line 340: "Anfrage zurückgenommen". Und "Anfrage abgelehnt" für das rejected.

lib/l10n/app_de.arb
```diff
@@ -330,6 +330,54 @@
...
+  "rideEventCancelledByDriverTitle": "Fahrt abgesagt!",
+  "rideEventCancelledByRiderTitle": "Mitfahrt abgesagt!",
+  "rideEventPendingMessage": "{username} hat eine Mitfahrt angefragt",
+  "@rideEventPendingMessage": {
```
fynsta, line 344: in the german arb file, you don't need these placeholders. They are taken from the english file

lib/l10n/app_en.arb
```diff
@@ -444,5 +444,54 @@
...
   "pageChatMessageBarHint": "Write a Message",
 
+  "pageHomeTitle": "Home",
+  "pageHomeSearchButton": "Search Ride",
+  "pageHomeCreateButton": "Create Drive",
+
+  "rideEventPendingTitle": "Ride requested!",
+  "rideEventApprovedTitle": "Ride approved!",
+  "rideEventRejectedTitle": "Ride rejected!",
+  "rideEventWithdrawnTitle": "Ride withdrawn!",
```
fynsta, line 454: also here "Request withdrawn" and "Request rejected"

lib/util/ride_event.dart
```diff
@@ -0,0 +1,113 @@
...
+    return jsonList.map((Map<String, dynamic> json) => RideEvent.fromJson(json)).toList();
+  }
+
+  @override
+  Map<String, dynamic> toJson() {
+    return <String, dynamic>{
+      'ride_id': rideId,
+      'category': category.index,
+      'read': read,
+    };
```
fynsta, line 44: the `toJsonForApi` method is missing.

lib/util/ride_event.dart
```diff
@@ -0,0 +1,113 @@
...
+    //custom rpc call to mark rideEvent as read, so the user does not need the write permission on the ride_events table
+    await SupabaseManager.supabaseClient.rpc('mark_ride_event_as_read', params: <String, dynamic>{'ride_event_id': id});
+  }
+
+  bool isForCurrentUser() {
+    final int profileId = SupabaseManager.getCurrentProfile()!.id!;
+    if (category == RideEventCategory.approved ||
+        category == RideEventCategory.cancelledByDriver ||
+        category == RideEventCategory.rejected) {
+      return ride!.rider!.id == profileId;
```
fynsta, line 58: riderId and driverId below would be simpler

test/models/ride_event_test.dart
```diff
@@ -0,0 +1,247 @@
...
+        riderId: user.id,
+      )),
+    );
+    final RideEvent rideEventWithUserNotRider = RideEventFactory().generateFake(
+      ride: NullableParameter(RideFactory().generateFake(
+        drive: NullableParameter(driveWithUserNotDriver),
+        riderId: user.id! + 1,
+      )),
+    );
+    setUp(() {
```
fynsta, line 40: please add some linebreaks, this is really hard to read

fynsta:
I like it a lot overall!

lib/home_page.dart
```diff
@@ -8,15 +24,288 @@ class HomePage extends StatefulWidget {
...
+    final Message message = Message.fromJson(data);
+    if (message.read == false) {
+      setState(() {
+        _items.insert(0, message);
+      });
+    }
+  }
+
+  Card _buildMessageWidget(Message message, BuildContext context) {
+    return Card(
```
fynsta, line 251: there should be a way to dismiss these cards by swiping. That probably requires a new database column (because a message should not be marked as read just because you swiped). But we can do that in a follow-up PR.

lib/home_page.dart
```diff
@@ -1,4 +1,20 @@
...
+import 'rides/pages/ride_detail_page.dart';
+import 'rides/pages/search_ride_page.dart';
+import 'util/chat/models/message.dart';
+import 'util/chat/pages/chat_page.dart';
+import 'util/model.dart';
+import 'util/parse_helper.dart';
+import 'util/ride_event.dart';
+import 'util/supabase.dart';
 
 class HomePage extends StatefulWidget {
```
Eddie-42, line 19: I had a null reference exception earlier today. I couldn't even open the app. Unfortunately, I can't reproduce it right now, and the stack trace was more than 1000 lines long, making it impossible for me to see the actual exception because the shell cuts off after 1000 lines (bad shell). But I think it was visible to everyone at the time, including AG 😬 Could you please check that every ! operator is used with merit? Otherwise, if this doesn't happen again, maybe the supabaseClient wasn't set and we'll rule it as a freak occurence

lib/home_page.dart
```diff
@@ -8,15 +24,288 @@ class HomePage extends StatefulWidget {
...
+        padding: const EdgeInsets.symmetric(
+          horizontal: 20,
+          vertical: 10,
+        ),
+        child: Column(
+          children: <Widget>[
+            const SizedBox(height: 30),
+            if (_fullyLoaded)
+              Expanded(
+                child: Container(
```
Eddie-42, line 144: I don't like this container. I think it could be worth a try putting the buttons at the top and make them scroll along with the events. If we want to keep the buttons in sight at all times, we can try a little CustomScrollView/PersistentHeader magic.

lib/home_page.dart
```diff
@@ -8,15 +24,288 @@ class HomePage extends StatefulWidget {
...
+                            const SizedBox(height: 16),
+                            Text(
+                              S.of(context).pageHomePageEmpty,
+                              style: Theme.of(context).textTheme.headline6,
+                            ),
+                            const SizedBox(height: 3),
+                          ],
+                        ),
+                ),
+              )
```
Eddie-42, line 177: Didn't we plan to put upcoming drives/rides here? Or is that going to be another PR?

lib/home_page.dart
```diff
@@ -8,15 +24,288 @@ class HomePage extends StatefulWidget {
...
+                  padding: const EdgeInsets.symmetric(vertical: 10, horizontal: 10),
+                  decoration: BoxDecoration(
+                    border: Border.all(color: Theme.of(context).primaryColor),
+                    borderRadius: const BorderRadius.all(Radius.circular(5.0)),
+                  ),
+                  child: _items.isNotEmpty
+                      ? ListView.separated(
+                          itemCount: _items.length,
+                          separatorBuilder: (BuildContext context, int index) {
+                            return const SizedBox(height: 12);
```
Eddie-42, line 155: I think the elements need way less separation, if any at all

lib/home_page.dart
MarkFalke: I think thats a great idea and I think we should do it in a different PR too.

lib/home_page.dart
```diff
@@ -8,15 +24,288 @@ class HomePage extends StatefulWidget {
...
+                          },
+                          itemBuilder: (BuildContext context, int index) {
+                            return _buildWidget(_items[index], context);
```
Eddie-42, line 158: Okay I realize this might not be an easy ask, but is it possible to group these by drives/rides? I think it's not very clear right now which notifications belong to which drives/rides (and chats are the worst offenders because they can spam the whole panel so that you have no overview at all anymore. Off the top of my head, I would have this functionality modelled after _shudder_ iOS notifications, any other opinions on this?

lib/home_page.dart
Eddie-42: If so, I would argue that we should definitely put upcoming drives/rides separately from events, in which case the whole page will become scrollable because there's not enough space for another panel right now and we should definitely find another solution for the "scrollable panel" solution if we don't want scrollables inside of scrollables

lib/home_page.dart
fynsta: He means the outline I think. And I don't know how it could look good in any way without the outline. Imo it's definitely needed.

lib/home_page.dart
fynsta: I like that, but I also think that should be a follow-up PR. This is complicated enough.

lib/home_page.dart
Eddie-42: There has to be a way to make this look better, and with the current design we will be running into the problems with upcoming trips I described below... But that can be addressed in another PR I suppose

lib/home_page.dart
MarkFalke: thanks for the hint

lib/home_page.dart
MarkFalke: i don't think that to panels are a good option, we can include them and pin them at the top.

lib/home_page.dart
```diff
@@ -8,15 +24,275 @@ class HomePage extends StatefulWidget {
...
+              )
+            else
+              const Center(
+                child: CircularProgressIndicator(),
+              ),
+            const SizedBox(height: 35),
+            Column(
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: <Widget>[
+                Hero(
```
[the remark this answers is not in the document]
MarkFalke: okay

lib/home_page.dart
```diff
@@ -8,15 +24,288 @@ class HomePage extends StatefulWidget {
...
+                    isforDrive ? DriveDetailPage(id: rideEvent.ride!.driveId) : RideDetailPage(id: rideEvent.rideId),
+              ),
+            );
+          },
+        ),
+      ),
+    );
+  }
+
+  Widget _buildWidget(Model model, BuildContext context) {
```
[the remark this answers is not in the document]
MarkFalke: 1) I changed the space to 1 between the cards is that okay or should i remove it completly?
2) Yes I thought that too, for the messages i don't think it's a big problem since it's clear for wich person the chat is. And for the events, i can add somekind of mini-timeline but i dont think that could replace the icon since it's way to big. What i could do is just add it in the subtitle?
3) yes i forgot that

lib/util/ride_event.dart
MarkFalke: changed it to rider/driver.isCurrentUser

test/models/ride_event_test.dart
```diff
@@ -0,0 +1,247 @@
...
+  setUp(() async {
+    MockServer.setProcessor(rideEventProcessor);
+  });
+
+  group('markAsRead', () {
+    test('marks rideEventas read', () async {
+      final RideEvent rideEvent = RideEventFactory().generateFake(
+        read: false,
+      );
+      await rideEvent.markAsRead();
```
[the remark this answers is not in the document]
MarkFalke: thanks for the hint

lib/home_page.dart
```diff
@@ -8,15 +26,275 @@ class HomePage extends StatefulWidget {
...
+      },
+    ).on(
+      RealtimeListenTypes.postgresChanges,
+      ChannelFilter(event: 'UPDATE', schema: 'public', table: 'messages', filter: 'sender_id=neq.$profileId'),
+      (dynamic payload, [dynamic ref]) {
+        if (pick(pick(payload, 'new').asMapOrEmpty(), 'read').asBoolOrThrow() == true) {
+          setState(() {
+            _items.removeWhere(
+              (Model element) =>
+                  element is Message && element.id == pick(pick(payload, 'new').asMapOrEmpty(), 'id').asIntOrThrow(),
```
fynsta, line 54: can't you just do `pick(payload, 'new', 'id')...` immediately?

fynsta:
I'm gonna review this properly again tomorrow or so, but @MarkFalke I have a few questions:
1) Do you have time the next few days? Would love to finish this toward Monday or Tuesday next week
2) I think we should write tests here before merging - now that everything else is tested already, tests should always be part of the PR.

lib/home_page.dart
```diff
@@ -1,22 +1,511 @@
...
+      _fullyLoaded = true;
+    });
+  }
+
   @override
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-        title: const Text('Home'),
+        title: Text('${S.of(context).hello} ${supabaseManager.currentProfile!.username} :)'),
```
Eddie-42, line 192:
```suggestion
        // Wave emoji
        title: Text('${S.of(context).hello} ${supabaseManager.currentProfile!.username} \u{1F44B}'),
```

Eddie-42: Also, I feel like the emoji should be part of the translation. Not because we want to use other emojis in different languages per se, but I feel like it would be cleaner and more future-proof (Right-to-left languages?)
So, I'd add a username parameter to the hello text in the arb file and include the emoji there

lib/l10n/app_en.arb
```diff
@@ -350,5 +335,30 @@
...
+  "pageHomeUpcomingDriveTomorrow": "Upcoming Drive tomorrow!",
+  "pageHomeUpcomingRideToday": "Upcoming Ride today!",
+  "pageHomeUpcomingRideTomorrow": "Upcoming Ride tomorrow!",
+  "rideEventPendingMessage": "{username} has requested a ride",
+  "rideEventApprovedMessage": "{username} has approved your ride",
+  "rideEventRejectedMessage": "{username} has rejected your ride",
+  "rideEventWithdrawnMessage": "{username} has withdrawn his ride",
+  "rideEventCancelledByDriverMessage": "{username} has cancelled his drive",
+  "rideEventCancelledByRiderMessage": "{username} has cancelled his ride",
+  "pageHomeUpcomingTripMessage": "from {start} to {end} at {startTime}",
```
Eddie-42, line 358: This is the wrong way around. @istzen @fynsta is sorry, but this is because the parameters aren't given in the arb file. So the order of parameters is apparently chosen alphabetically. I think we have to include the lengthy placeholder syntax for multiple parameters

lib/l10n/app_en.arb
```diff
@@ -350,5 +335,30 @@
...
+  "rideEventCancelledByDriverTitle": "Drive cancelled!",
+  "rideEventCancelledByRiderTitle": "Ride cancelled!",
+  "pageHomeUpcomingDriveToday": "Upcoming Drive today!",
+  "pageHomeUpcomingDriveTomorrow": "Upcoming Drive tomorrow!",
+  "pageHomeUpcomingRideToday": "Upcoming Ride today!",
+  "pageHomeUpcomingRideTomorrow": "Upcoming Ride tomorrow!",
+  "rideEventPendingMessage": "{username} has requested a ride",
+  "rideEventApprovedMessage": "{username} has approved your ride",
+  "rideEventRejectedMessage": "{username} has rejected your ride",
+  "rideEventWithdrawnMessage": "{username} has withdrawn his ride",
```
Eddie-42, line 355:
```suggestion
  "rideEventWithdrawnMessage": "{username} has withdrawn his request",
```

lib/l10n/app_en.arb
```diff
@@ -350,5 +335,30 @@
...
+  "rideEventCancelledByDriverMessage": "{username} has cancelled his drive",
+  "rideEventCancelledByRiderMessage": "{username} has cancelled his ride",
+  "pageHomeUpcomingTripMessage": "from {start} to {end} at {startTime}",
+  "pageHomePageEmpty": "No new messages",
+
+  "hello": "Hello",
+
   "appName": "Motis Ride-Sharing App"
 }
```
Eddie-42, line 364: Has there always been no newline at the end of the arb file?

lib/home_page.dart
```diff
@@ -1,22 +1,511 @@
...
+          _upcomingTripsCount--;
+        }
+      });
+    }
+  }
+
+  Card _buildMessageWidget(Message message) {
+    return Card(
+      child: InkWell(
+        child: Dismissible(
```
Eddie-42, line 397: I think the whole card should be swiped away, not just the text.

fynsta (approved):

fynsta:
I'm gonna approve this because I finally want to see it merged :D Any remaining stuff we can clean up later, but this is open for way too long

Showcas (approved):

istzen (approved):

PR #91, opened 2023-01-25
Chat Tests

istzen:
tests for chat and everything that belongs to it. That means we have now 100% coverage for ride_detail and drive_detail. And also for the chat and everything that belongs to it we have full coverage except some lines that depend on the stream i can not currently cover.


lib/drives/pages/drive_detail_page.dart
```diff
@@ -236,32 +236,7 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
-                    ),
-                    showBadge: _drive!.getUnreadMessagesCount() != 0,
-                    position: BadgePosition.topEnd(top: -12),
-                    child: const Icon(Icons.chat),
-                  ),
-                  tooltip: S.of(context).openChat,
-                ),
-              ]
-            : null,
+        actions: <Widget>[buildChatButton()],
```
fynsta, line 239: ah was that because of a merge conflict? I remember Simon and I built that method to handle a null drive

test/widgets/chat/chat_page_test.dart
```diff
@@ -0,0 +1,176 @@
...
+  testWidgets('shows ProfileWidget of given Profile', (WidgetTester tester) async {
+    await pumpMaterial(tester, ChatPage(profile: profile, chatId: chatId));
+    final Finder profileWidgetFinder = find.byType(ProfileWidget);
+    expect(profileWidgetFinder, findsOneWidget);
+    final ProfileWidget profileWidget = tester.widget(profileWidgetFinder);
+    expect(profileWidget.profile, profile);
+  });
+
+  group('Messagebar', () {
+    setUp(() => whenRequest(processor).thenReturnJson([]));
```
fynsta, line 46: can you use a few more line breaks after setUp and between the testWidgets themselves :) Would be a lot easiert to read

test/widgets/chat/chat_page_test.dart
```diff
@@ -0,0 +1,176 @@
...
+      await pumpMaterial(tester, ChatPage(profile: profile, chatId: chatId));
+      await tester.pump();
+      final Finder chatBubble = find.byType(ChatBubble);
+      expect(chatBubble, findsOneWidget);
+      expect(find.descendant(of: chatBubble, matching: find.byIcon(Icons.done_all)), findsNothing);
+    });
+    testWidgets('shows tail iff it should be shown', (WidgetTester tester) async {
+      final Profile currentProfile = ProfileFactory().generateFake(id: 2);
+      SupabaseManager.setCurrentProfile(currentProfile);
+      final List<Map<String, dynamic>> messages = [
```
fynsta, line 123: Not sure I like this test. Maybe the messages should each get a descriptive name that explains why they are not shown?

test/widgets/chat/drive_chat_page_test.dart
```diff
@@ -0,0 +1,333 @@
...
+        urlMatcher: equals('/rest/v1/messages?select=%2A&order=created_at.desc.nullslast'),
+      );
+      final List<RealtimeChannel> subscription = SupabaseManager.supabaseClient.getChannels();
+      expect(subscription.length, 0);
+    });
+
+    testWidgets('not when Drive has rides without active chat', (WidgetTester tester) async {
+      drive = DriveFactory().generateFake(
+        driverId: profile.id,
+        rides: [RideFactory().generateFake(status: RideStatus.pending)],
```
fynsta, line 71: maybe also add a ride with another status to this array, just to make the point more clear.

test/widgets/chat/drive_chat_page_test.dart
```diff
@@ -0,0 +1,333 @@
...
+  });
+
+  group('Has Stream Subscription', () {
+    setUp(() async {
+      whenRequest(processor).thenReturnJson([]);
+    });
+    testWidgets('when Drive has rides with active chat', (WidgetTester tester) async {
+      drive = DriveFactory().generateFake(
+        driverId: profile.id,
+        rides: [RideFactory().generateFake(chatId: 1, status: RideStatus.approved)],
```
fynsta, line 40: we should probably add another ride without chat here to have a more complete test case

test/widgets/chat/drive_chat_page_test.dart
```diff
@@ -0,0 +1,333 @@
...
+      final List<RealtimeChannel> subscription = SupabaseManager.supabaseClient.getChannels();
+      expect(subscription.length, 0);
+    });
+  });
+
+  testWidgets('Shows chats from Rides that have a active Chat', (WidgetTester tester) async {
+    drive = DriveFactory().generateFake(
+      driverId: profile.id,
+      rides: [
+        RideFactory().generateFake(chatId: 1, status: RideStatus.approved),
```
fynsta, line 88: why do these chatIds need to be hard-coded? You are not even using them later. But generally, I think we should hard-code way less of these ideas anyways.

test/widgets/drive_detail_page_test.dart
```diff
@@ -141,8 +142,33 @@ void main() {
...
+      await pumpMaterial(tester, DriveDetailPage.fromDrive(drive));
 
-      // Add navigating to chat here
+      await tester.pump();
+      await pumpMaterial(tester, DriveDetailPage.fromDrive(drive));
+      await tester.pump();
+      await tester.tap(find.byKey(Key('chatPageButton${drive.rides!.last.id}Start')));
+      await tester.pumpAndSettle();
+      await tester.pump();
+      await tester.pumpAndSettle();
```
fynsta, line 165: this whole section is really weird. Way to many pumps for my taste lol. Probably forgot to clean up after testing around?

test/widgets/drive_detail_page_test.dart
```diff
@@ -255,13 +281,13 @@ void main() {
...
-    testWidgets('Can navigate to drive chat page', skip: true, (WidgetTester tester) async {
+    testWidgets('Can navigate to drive chat page', (WidgetTester tester) async {
       await pumpMaterial(tester, DriveDetailPage.fromDrive(drive));
       await tester.pump();
 
       await tester.tap(find.byKey(const Key('driveChatButton')));
-      await tester.pumpAndSettle();
-
+      await tester.pump();
+      await tester.pump();
```
fynsta, line 290: multiple pumps? why is that needed?

lib/drives/pages/drive_detail_page.dart
istzen: yes i guess i messed that up while rebasing but now it's back again.

test/widgets/chat/drive_chat_page_test.dart
istzen: yes i can do that i thought that the point is clear enough because the activeChat method is there and tested

test/widgets/drive_detail_page_test.dart
istzen: yes thanks that was the problem with the Hero xD

test/widgets/drive_detail_page_test.dart
istzen: because of the stream

test/widgets/chat/chat_page_test.dart
istzen: did that

test/widgets/chat/drive_chat_page_test.dart
istzen: because otherwise i could have the same ids then the test would fail.

test/widgets/chat/chat_page_test.dart
istzen: i added something not exactly sure what to ride

istzen: *write xD

test/widgets/chat/drive_chat_page_test.dart
fynsta: The factory generates ids as a random number between 1 and 10000000. I don't think that's likely :D

test/widgets/chat/chat_page_test.dart
fynsta: Yeah I would have done
```
messageForCurrentUser = ...;
otherMessageForCurrentUser = ...;
newestMessageForCurrentUser = ...;
...
messages = [messageForCurrentUser, otherMessageForCurrentUser, newestMessageForCurrentUser, ...]

But I guess your way is also fine.

File: test/widgets/drive_detail_page_test.dart (hunk 8 above)
Comment (fynsta) [reply test/widgets/drive_detail_page_test.dart:299]: Good to know! But `pumpAndSettle` does not work?

File: test/widgets/chat/drive_chat_page_test.dart (hunk 6 above)
Comment (istzen) [reply test/widgets/chat/drive_chat_page_test.dart:99]: Yes but still possible and then you have a test that fails and don't know why

Comment (fynsta) [reply test/widgets/chat/drive_chat_page_test.dart:99]: You would do the test locally and see that it was just fluky... also these odds are insanely low if you calculate it out

Comment (fynsta) [reply test/widgets/chat/drive_chat_page_test.dart:99]: and we can easily add one more zero if the number is not high enough lol

Comment (istzen) [reply test/widgets/chat/drive_chat_page_test.dart:99]: but then the possibility would still be there, i just don't see why that's nessecary.

File: test/widgets/chat/chat_page_test.dart [hunk 9]
```diff
@@ -0,0 +1,202 @@
...
+  testWidgets('shows ProfileWidget of given Profile', (WidgetTester tester) async {
+    await pumpMaterial(tester, ChatPage(profile: profile, chatId: chatId));
+
+    final Finder profileWidgetFinder = find.byType(ProfileWidget);
+    expect(profileWidgetFinder, findsOneWidget);
+    final ProfileWidget profileWidget = tester.widget(profileWidgetFinder);
+    expect(profileWidget.profile, profile);
+  });
+
+  group('Messagebar', () {
```
Eddie-42, line 48:
```suggestion
  group('MessageBar', () {
```

test/widgets/chat/chat_page_test.dart
```diff
@@ -0,0 +1,202 @@
...
+      ]);
+      await pumpMaterial(tester, ChatPage(profile: profile, chatId: chatId));
+      await tester.pump();
+
+      final Finder chatBubble = find.byType(ChatBubble);
+      expect(chatBubble, findsOneWidget);
+      expect(find.descendant(of: chatBubble, matching: find.byIcon(Icons.done_all)), findsNothing);
+    });
+
+    testWidgets('shows tail iff it should be shown', (WidgetTester tester) async {
```
Eddie-42, line 138:
```suggestion
    testWidgets('shows tail when it should be shown', (WidgetTester tester) async {
```

test/widgets/chat/drive_chat_page_test.dart
Eddie-42: I have to take Fynn's side here. Tests are never going to be 100.0% accurate, and I think going for the last literal 0.0000001% is the wrong incentive here (for comparison, the most advanced autonomous vehicles have an injury collision avoidance performance of 99.9999%, which would make them less safe than this test). That comes at the cost of a little wasted space where the reader has to spend some time thinking about why the IDs were specified, which is not the point of the test. It makes the test less focused, less readable and less clean

istzen: okay I'm convinced by the argument with the autonomous vehicle.

test/widgets/chat/chat_page_test.dart
istzen: i was going for if and only if but i can change it

fynsta: Mathematically it's correct with "iff" :D But maybe offputting for non-mathematicians (not just Simon, but everyone else who might read that code)

Showcas (approved):

fynsta (approved):

PR #92, opened 2023-01-27
Test welcome

fynsta:
This adds 100% test coverage for the whole "Welcome"-section of our app (Welcome, Register, Login, ForgotPassword, ResetPassword, AfterRegistration). Got a little bigger than I expected, but it's finally done. I also reformatted a little bit of the existing code in search for better readability.

Most of the tests are pretty straightforward, but sometimes I had to look deep into the supabase/GoTrue code to figure out the correct response to mock. If you have any questions, please don't hesitate to ask.

Other important general changes:
- The `thenReturnJson` method now has an optional `statusCode` parameter so that we can mock error responses. This required to rewrite the processor a little bit, as you can see for yourself. This also means that the `thenReturn` method **will not work anymore**.
- The `pumpMaterial` now has an optional `navigatorObserver` parameter. You can add a MockNavigatorObserver here to check whether "pop", "push" etc. were actually called (often not needed, but can definitely be helpful).
- I added a `pumpForm` helper because it is a little bit more complicated to unit test a form field (you need a form with form key, and also a `Material` object like a Scaffold around it)
- Moved all mock-related stuff to a `mocks` directory. This might destroy your branches, so you will have to be careful and clean that up after rebasing.


.github/workflows/flutter-build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: flutter pub get
 
       - name: Verify formatting
-        run: flutter format --output=none --set-exit-if-changed --line-length=120 .
+        run: flutter format --output=none --set-exit-if-changed --line-length=120 $(find lib test -name "*.dart" -not -name "*.mocks.dart")
```
fynsta, line 31: If you are wondering: This was necessary because the mock files are not formatted correctly (with our bigger line length). So we need to exclude them from the formatting in our CI.

istzen:
not sure if i like our trend to put lot's of different stuff in one PR.

fynsta:
> not sure if i like our trend to put lot's of different stuff in one PR.

Do you mean the welcome section should have been split into a PR for each page? Maybe, but some of those would have been minuscule, and most of these are so similar that there is a lot of overlap in style as well as the actual tests.

Or do you mean the other changes to the request processor etc.? Because those were definitely needed to make the things in the PR testable.

test/widgets/after_registration_page_test.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import '../util/pump_material.dart';
+
+void main() {
+  setUpAll(() async {
+    SupabaseManager.setCurrentProfile(null);
+  });
+
+  group('WelcomePage', () {
+    testWidgets('Shows the page', (WidgetTester tester) async {
```
istzen, line 16: is this test really necessary or did you mean `'Shows the image'`.

test/widgets/login_page_test.dart
```diff
@@ -0,0 +1,235 @@
...
+      of: find.byKey(const Key('loginEmailField')),
+      matching: find.byType(TextFormField),
+    );
+    final Finder passwordFieldFinder = find.descendant(
+      of: find.byKey(const Key('loginPasswordField')),
+      matching: find.byType(TextFormField),
+    );
+    final Finder loginButtonFinder = find.byType(LoadingButton);
+
+    testWidgets('Shows the login page', (WidgetTester tester) async {
```
istzen, line 36:
```suggestion
    testWidgets('Shows the login form', (WidgetTester tester) async {
```

test/widgets/login_page_test.dart
```diff
@@ -0,0 +1,235 @@
...
+
+      formState.formKey.currentState!.validate();
+      await tester.pumpAndSettle();
+      // Empty, so error
+      expect(passwordField.hasError, isTrue);
+
+      await tester.enterText(passwordFieldFinder, 'abcdefg');
+      formState.formKey.currentState!.validate();
+      await tester.pumpAndSettle();
+      expect(passwordField.hasError, isFalse);
```
istzen, line 91: shouldn't that password be invalid since there is no uppercase letter and no special character?

test/widgets/after_registration_page_test.dart
fynsta: Now I just want to test that it shows _something_ and does not crash. The page is literally just an image and some text

test/widgets/login_page_test.dart
fynsta: We don't validate the login page as strictly. It does not matter to us if the password is not following our rules... In that case we can just show it is wrong anyways.

That's why the test description also mentions "(not as strictly)"

test/widgets/login_page_test.dart
fynsta: I mean, yeah sure but it also shows the page

test/widgets/register_page_test.dart
```diff
@@ -0,0 +1,322 @@
...
+    testWidgets('Shows the register page', (WidgetTester tester) async {
+      await pumpMaterial(tester, const RegisterPage());
+
+      final Finder registerFormFinder = find.byType(RegisterForm);
+      final RegisterFormState formState = tester.state(registerFormFinder);
+
+      expect(formState.buttonState, ButtonState.idle);
+    });
+
+    testWidgets('Validates the email', (WidgetTester tester) async {
```
istzen, line 51: can we refactor the test of the emailField since the code is the same for register and login, same for Password field.

test/widgets/register_page_test.dart
```diff
@@ -0,0 +1,322 @@
...
+      expect(emailField.hasError, isTrue);
+
+      await tester.enterText(emailFieldFinder, email);
+      formState.formKey.currentState!.validate();
+      await tester.pumpAndSettle();
+      // Real email, so no error
+      expect(emailField.hasError, isFalse);
+    });
+
+    testWidgets('Validates the username', (WidgetTester tester) async {
```
istzen, line 79: please add a case where the username is to long.

istzen:
> > not sure if i like our trend to put lot's of different stuff in one PR.
> 
> Do you mean the welcome section should have been split into a PR for each page? Maybe, but some of those would have been minuscule, and most of these are so similar that there is a lot of overlap in style as well as the actual tests.
> 
> Or do you mean the other changes to the request processor etc.? Because those were definitely needed to make the things in the PR testable.

the second one. I understand that this changes needed to be done for this PR but i think it would have been better to divide them into 2.

test/widgets/after_registration_page_test.dart
istzen: ah okay i understand.

test/widgets/register_page_test.dart
```diff
@@ -0,0 +1,322 @@
...
+      // No special, so error
+      expect(passwordField.hasError, isTrue);
+
+      await tester.enterText(passwordFieldFinder, password);
+      formState.formKey.currentState!.validate();
+      await tester.pumpAndSettle();
+      expect(passwordField.hasError, isFalse);
+    });
+
+    testWidgets('Validates the password confirmation', (WidgetTester tester) async {
```
istzen, line 160: what happens if they are the same but both  do not meet the requirements  just wondering?

test/widgets/register_page_test.dart
```diff
@@ -0,0 +1,322 @@
...
+      testWidgets('Fails if form is not filled validly', (WidgetTester tester) async {
+        await pumpMaterial(tester, const RegisterPage());
+
+        await fillForm(tester, valid: false);
+
+        await tester.tap(registerButtonFinder);
+        await tester.pump();
+
+        final FormFieldState passwordConfirmationField = tester.state(passwordConfirmFieldFinder);
+        expect(passwordConfirmationField.hasError, isTrue);
```
istzen, line 238: can you add a verifyNever call here?

test/widgets/reset_password_page_test.dart
```diff
@@ -0,0 +1,192 @@
...
+      // No special, so error
+      expect(passwordField.hasError, isTrue);
+
+      await tester.enterText(passwordFieldFinder, password);
+      formState.formKey.currentState!.validate();
+      await tester.pumpAndSettle();
+      expect(passwordField.hasError, isFalse);
+    });
+
+    testWidgets('Validates the password confirmation', (WidgetTester tester) async {
```
istzen, line 127: again can you refactor this

test/widgets/register_page_test.dart
fynsta: I know, I thought the same thing... But I don't actually know how you would do that. I can try to look into it. Probably just move those tests to EmailField and then mock it some way

test/widgets/register_page_test.dart
fynsta: Ah actually we forgot to add the restriction in the register field. I will add and test it.

test/widgets/register_page_test.dart
fynsta: I guess both of them will show an error, but I don't think that's a relevant case right? It's covered by these two test cases.

test/widgets/reset_password_page_test.dart
fynsta: No here I really don't know how to refactor the tests. It's not just one widget, it's the combination of multiple widgets.

The only way to do this would be to move "PasswordField + PasswordConfirmationFIeld" into a new widget, that could then be tested. But as long as the original code is duplicated, the test needs to be duplicated as well.

fynsta:
> the second one. I understand that this changes needed to be done for this PR but i think it would have been better to divide them into 2.

But the only way to test the new request processor syntax was to add tests that use it. I don't really see how that could have been done otherwise?

fynsta:
The only thing I will concede is that moving the mocks into their own directory should have been done in another PR. If you want, I can revert that part.

test/widgets/register_page_test.dart
fynsta: Wrote unit tests and removed the "Validates the email" test cases. It's not 100% the same thing, but should be close enough.

Eddie-42 (approved):

Showcas (approved):

istzen:
> The only thing I will concede is that moving the mocks into their own directory should have been done in another PR. If you want, I can revert that part.

no don't worry about it, i don't think it's a big deal. What i meant was to open one PR for the editing for the Processor and one that works based on that branch for the tests. Then the first part could have been merged more quickly and could have been used everywhere else already.

istzen (approved):

PR #93, opened 2023-01-27
Cards test

istzen:
Tests for Cards, note that the coverage is not that good since we have no method to test the color if I am correct, otherwise please let me know then I'll add tests to check for the status color.

Add pumpScaffold method for all widgets that don't have a scaffold themselves.

### DIFF
--- a/lib/util/trip/pending_ride_card.dart
+++ b/lib/util/trip/pending_ride_card.dart
@@ -133,7 +133,7 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
                   context,
                   S.of(context).cardPendingRideApproveDialogSuccessSnackbar,
                   durationType: SnackBarDurationType.medium,
-                  key: const Key('approveSuccesSnackbar'),
+                  key: const Key('approveSuccessSnackbar'),
                 );
               } else {
                 Navigator.of(dialogContext).pop();
--- a/lib/util/trip/pending_ride_card.dart
+++ b/lib/util/trip/pending_ride_card.dart
@@ -40,7 +40,7 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
 
   @override
   Widget buildTopRight() {
-    return Text('+${localeManager.formatDuration(extraTime, shouldPadHours: false)}');
+    return Text(key: const Key('extraTime'), '+${localeManager.formatDuration(extraTime, shouldPadHours: false)}');
   }
 
   @override
@@ -55,7 +55,7 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
   Widget buildBottomRight() {
     return Padding(
       padding: const EdgeInsets.all(8.0),
-      child: Text('${_ride.price}€'),
+      child: Text(key: const Key('price'), '${_ride.price}€'),
     );
   }
 
@@ -64,11 +64,13 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
     return ButtonBar(
       children: <Widget>[
         IconButton(
+          key: const Key('approveButton'),
           onPressed: () => showApproveDialog(context),
           icon: const Icon(Icons.check_circle_outline, color: Colors.green, size: 50.0),
           tooltip: S.of(context).approve,
         ),
         IconButton(
+          key: const Key('rejectButton'),
           onPressed: () => showRejectDialog(context),
           icon: const Icon(Icons.cancel_outlined, color: Colors.red, size: 50.0),
           tooltip: S.of(context).reject,
@@ -115,10 +117,12 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
         content: Text(S.of(context).cardPendingRideApproveDialogMessage),
         actions: <Widget>[
           TextButton(
+            key: const Key('approveCancelButton'),
             child: Text(S.of(context).no),
             onPressed: () => Navigator.of(dialogContext).pop(),
           ),
           TextButton(
+            key: const Key('approveConfirmButton'),
             child: Text(S.of(context).yes),
             onPressed: () {
               // check if there are enough seats available
@@ -129,11 +133,13 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
                   context,
                   S.of(context).cardPendingRideApproveDialogSuccessSnackbar,
                   durationType: SnackBarDurationType.medium,
+                  key: const Key('approveSuccesSnackbar'),
                 );
               } else {
                 Navigator.of(dialogContext).pop();
                 showSnackBar(
                   context,
+                  key: const Key('approveErrorSnackbar'),
                   S.of(context).cardPendingRideApproveDialogErrorSnackbar,
                   durationType: SnackBarDurationType.medium,
                 );
@@ -153,16 +159,19 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
         content: Text(S.of(context).cardPendingRideRejectDialogMessage),
         actions: <Widget>[
           TextButton(
+            key: const Key('rejectCancelButton'),
             child: Text(S.of(context).no),
             onPressed: () => Navigator.of(context).pop(),
           ),
           TextButton(
+            key: const Key('rejectConfirmButton'),
             child: Text(S.of(context).yes),
             onPressed: () {
               rejectRide();
               Navigator.of(context).pop();
               showSnackBar(
                 context,
+                key: const Key('rejectSuccessSnackbar'),
                 S.of(context).cardPendingRideRejectDialogSuccessSnackBar,
                 durationType: SnackBarDurationType.medium,
               );
--- a/lib/util/trip/trip_card.dart
+++ b/lib/util/trip/trip_card.dart
@@ -71,6 +71,7 @@ abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends Stat
               mainAxisAlignment: MainAxisAlignment.spaceBetween,
               children: <Widget>[
                 Text(
+                  key: const Key('start'),
                   '${localeManager.formatTime(trip.startTime)}  ${trip.start}',
                   style: const TextStyle(fontWeight: FontWeight.bold, fontSize: 16),
                 ),
@@ -115,6 +116,7 @@ abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends Stat
               mainAxisAlignment: MainAxisAlignment.spaceBetween,
               children: <Widget>[
                 Text(
+                  key: const Key('end'),
                   '${localeManager.formatTime(trip.endTime)}  ${trip.end}',
                   style: const TextStyle(fontWeight: FontWeight.bold, fontSize: 16),
                 ),
--- a/lib/util/trip/trip_card.dart
+++ b/lib/util/trip/trip_card.dart
@@ -182,9 +182,7 @@ abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends Stat
     return const SizedBox();
   }
 
-  Widget buildRightSide() {
-    return const SizedBox();
-  }
+  Widget buildRightSide();
 
   Color pickStatusColor() {
     return Theme.of(context).cardColor;
--- a/lib/util/trip/trip_card.dart
+++ b/lib/util/trip/trip_card.dart
@@ -182,7 +182,9 @@ abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends Stat
     return const SizedBox();
   }
 
-  Widget buildRightSide();
+  Widget buildRightSide() {
+    return const SizedBox();
+  }
 
   Color pickStatusColor() {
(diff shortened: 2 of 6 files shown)


lib/util/trip/trip_card.dart
```diff
@@ -180,9 +182,7 @@ abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends Stat
     return const SizedBox();
   }
 
-  Widget buildRightSide() {
-    return const SizedBox();
-  }
+  Widget buildRightSide();
```
fynsta, line 185: wait why did you change this? some other card might not want to implement this

istzen: Because it's overwritten in every subclass but i can change it Back.

fynsta: Yeah but that's not guaranteed right?

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+
+  setUp(() {
+    reset(processor);
+  });
+
+  //since the code is the same for RideCard DriveCard and PendingRideCard, we only test DriveCard
+  testWidgets('TripCard shows the correct information', (WidgetTester tester) async {
+    drive = DriveFactory().generateFake();
+    whenRequest(processor).thenReturnJson(drive.toJsonForApi());
+    await pumpMaterial(tester, DriveCard(drive, key: const Key('driveCard')));
```
fynsta, line 45: what's the point of passing that key?

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+      final Finder profileWidgetFinder = find.byType(ProfileWidget);
+      expect(profileWidgetFinder, findsOneWidget);
+      final ProfileWidget profileWidget = tester.widget(profileWidgetFinder);
+      expect(profileWidget.profile, ride.rider);
+
+      //Extra time is shown
+      expect(find.byKey(const Key('extraTime')), findsOneWidget);
+
+      //Seats indicator is shown
+      final Finder seatsIndicator = find.byType(IconWidget);
```
fynsta, line 88: Maybe better to check for an actual `SeatsIndicator` object.

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+      expect(find.byKey(const Key('approveButton')), findsOneWidget);
+
+      //Reject button is shown
+      expect(find.byKey(const Key('rejectButton')), findsOneWidget);
+    });
+
+    group('approve Ride', () {
+      Future<void> openApproveDialog(WidgetTester tester) async {
+        //need scaffold for dialog
+        await pumpMaterial(tester, Scaffold(body: PendingRideCard(ride, reloadPage: () {}, drive: drive)));
```
fynsta, line 105: tbh maybe we should just add a Scaffold by default to `pumpMaterial`

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+        await tester.tap(approveButton);
+        //load dialog
+        await tester.pumpAndSettle();
+      }
+
+      testWidgets('can approve Ride if possible', (WidgetTester tester) async {
+        await openApproveDialog(tester);
+
+        //confirm dialog
+        final Finder confimrButton = find.byKey(const Key('approveConfirmButton'));
```
fynsta, line 117: typo

lib/util/trip/pending_ride_card.dart
```diff
@@ -126,6 +130,7 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
                 Navigator.of(dialogContext).pop();
                 ScaffoldMessenger.of(context).showSnackBar(
                   SnackBar(
+                    key: const Key('approveSuccesSnackbar'),
```
fynsta, line 133: success is spelled with 2x "s"

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+          ProfileFeatureFactory().generateFake(rank: 1, feature: Feature.accessible),
+          ProfileFeatureFactory().generateFake(rank: 2, feature: Feature.noSmoking),
+          ProfileFeatureFactory().generateFake(rank: 3, feature: Feature.noVaping),
+          ProfileFeatureFactory().generateFake(rank: 4, feature: Feature.noPetsAllowed),
+        ]);
+
+        expect(find.byIcon(Icons.accessibility), findsOneWidget);
+        expect(find.byIcon(Icons.smoke_free), findsOneWidget);
+        expect(find.byIcon(Icons.vape_free), findsOneWidget);
+
```
fynsta, line 269: there should be a test that there is no `pets` icon

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+        ///7 Icons are always shown
+        expect(find.byType(Icon), findsNWidgets(10));
```
fynsta, line 271: I would remove this, that line would always have to be changed when we change completely unrelated parts of the card. Probably better to give the feature row a key and check that there are 3 icon descendants

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+        await tester.pumpAndSettle();
+
+        expect(find.byKey(const Key('rejectSuccessSnackbar')), findsNothing);
+        verifyRequestNever(processor, urlMatcher: equals('/rest/v1/rpc/reject_ride'));
+      });
+    });
+  });
+
+  group('RideCard', () {
+    Future<void> loadRideCard(WidgetTester tester, List<ProfileFeature>? features) async {
```
fynsta, line 225: features should probably not be a positional parameter, it should be named

fynsta:
Regarding testing color. I just looked up how you do that.

First, you need to get BuildContext of the app. That can be done with
`final BuildContext context = tester.element(find.byType(Container).first);`

Then you find whatever element you want to get the color of. For example
`final Finder accessibleIcon = find.byIcon(Icons.accessibility);`

And finally you query the color attribute of the corresponding widget:
`expect(tester.widget<Icon>(accessibleIcon).color, Theme.of(context).colorScheme.primary);`

lib/util/trip/trip_card.dart
Eddie-42: I think the change is fine. It serves as a reminder that TripCard is, indeed, abstract, and if an extending class doesn't want anything there, it should implement that itself. This just forces us to think about it

lib/util/trip/trip_card.dart
```diff
@@ -180,9 +182,7 @@ abstract class TripCardState<T extends Trip, U extends TripCard<T>> extends Stat
     return const SizedBox();
```
Eddie-42, line 182: IMO, we should do the same thing as buildRightSide here then

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+          methodMatcher: equals('POST'),
+          bodyMatcher: equals({'ride_id': ride.id}),
+        ).called(1);
+      });
+
+      testWidgets('can not approve Ride if not possible', (WidgetTester tester) async {
+        drive = DriveFactory().generateFake(rides: [ride], seats: 1);
+        ride = RideFactory().generateFake(
+          status: RideStatus.pending,
+          rider: NullableParameter(ProfileFactory().generateFake(id: 1)),
```
Eddie-42, line 137: Why is the ID needed?

test/widgets/cards_test.dart
```diff
@@ -0,0 +1,344 @@
...
+        expect(find.byKey(const Key('approveSuccesSnackbar')), findsOneWidget);
+        verifyRequest(
+          processor,
+          urlMatcher: equals('/rest/v1/rpc/approve_ride'),
+          methodMatcher: equals('POST'),
+          bodyMatcher: equals({'ride_id': ride.id}),
+        ).called(1);
+      });
+
+      testWidgets('can not approve Ride if not possible', (WidgetTester tester) async {
```
Eddie-42, line 133: Maybe include something along the lines of "too many seats" instead of "not possible" for more clarity when just reading the error messages

test/widgets/cards_test.dart
Eddie-42: In the next test, too

lib/util/trip/trip_card.dart
fynsta: No? It's much better if you just fill the spaces that you want. The rest should obviously always be empty boxes, so that should also be the default given in `TripCard`.

Eddie-42: I see two different philosophies here. I like it explicit, you like it implicit. That's fine, but I like it explicit

istzen: okay so what do i do now? should i change it back or not? what do the others think @Showcas @MarkFalke

test/widgets/cards_test.dart
istzen: good question xD

test/widgets/cards_test.dart
istzen: yes good idea thansk

istzen: actually i just looked into the code, there is nor `SeatIndicator` object there, but a `IconWidget` with icons.chair that's why i test it that way. I can however change it to a `SeatIndicator` not sure if that's the best way. 
@Showcas can you explain why there is a `IconWidget` used here and no `SeatIndicator`?

test/widgets/cards_test.dart
istzen: not sure since for the pages there already is a scaffold i think then it would be better to just add either a parameter `hasScaffold` or something or just use two methods

test/widgets/cards_test.dart
istzen: thanks

lib/util/trip/pending_ride_card.dart
istzen: thanks

test/widgets/cards_test.dart
istzen: thanks i had that not sure why i removed it xD

test/widgets/cards_test.dart
istzen: i don't understad what you mean exactly

test/widgets/cards_test.dart
istzen: did that

test/widgets/cards_test.dart
istzen: which id? xD

test/widgets/cards_test.dart
istzen: yes that's a very good idea thanks

test/widgets/cards_test.dart
istzen: i added some more xD

istzen:
@fynsta thank you very much for the help with the colors i added test for them also so the coverage is over 90% now for each card. (only for the did updateWidget method it's missing).

test/widgets/cards_test.dart
fynsta: Yes but when we write Widget tests that are not pages, we will always need this

test/widgets/cards_test.dart
fynsta: I think Simon and I forgot to change it there when we added `SeatIndicator`. Just change it as part of this PR, that should have been like that in the first place.

test/widgets/cards_test.dart
fynsta: Or a `pumpScaffold` helper?

lib/util/trip/trip_card.dart
fynsta: In any case, you need to do something. Either you remove all of these "build<Position>" methods and add the stubs to the descendants. Or you just remove this change and implement it in this class again.

test/widgets/cards_test.dart
istzen: yes that's what i meant. should i add it in this PR?

lib/util/trip/trip_card.dart
istzen: ok i guess i just change it back then xD

test/widgets/cards_test.dart
istzen: yes but they do different things. `IconWidget` shows if the count is bigger than two only one icon and the count and `SeatIndicator` shows just the number of icons and i think the first one is better for a card

test/widgets/cards_test.dart
fynsta: Let's do it :D

test/widgets/cards_test.dart
fynsta: Ah I see. In this case maybe we should still do this via the SeatIndicator with an optional boolean parameter (and that can just call IconWidget when it is true). But should be done in later refactorings of course.

fynsta (approved):

Eddie-42 (approved):

MarkFalke (approved):

Showcas (approved):

PR #94, opened 2023-01-27
Fix editing profile features

fynsta:
When you clicked on the button twice, then the features would be saved twice (and it would pop twice, which was even weirder). So the button needs to be disabled after the first click.

(@istzen are you happy with such a small PR? :D)


Eddie-42 (approved):

Showcas (approved):

istzen:
> When you clicked on the button twice, then the features would be saved twice (and it would pop twice, which was even weirder). So the button needs to be disabled after the first click.
> 
> (@istzen are you happy with such a small PR? :D)

yes perfect xD

istzen (approved):

PR #95, opened 2023-01-28
Test main

fynsta:
Some tests for `main.dart`. I don't know how to test the `main` method itself, so that is probably one of the 10% we cannot test.

Also missing is the `onError` part. I will do that when we merge test-welcome because I need to return an error via the api for this, of course.

### DIFF
--- a/test/widgets/main_test.dart
+++ b/test/widgets/main_test.dart
@@ -9,10 +9,10 @@ import 'package:motis_mitfahr_app/welcome/pages/reset_password_page.dart';
 import 'package:motis_mitfahr_app/welcome/pages/welcome_page.dart';
 
 import '../util/factories/profile_factory.dart';
-import '../util/mock_server.dart';
+import '../util/mocks/mock_server.dart';
+import '../util/mocks/request_processor.dart';
+import '../util/mocks/request_processor.mocks.dart';
 import '../util/pump_material.dart';
-import '../util/request_processor.dart';
-import '../util/request_processor.mocks.dart';
 
 void main() {
   final MockRequestProcessor processor = MockRequestProcessor();
--- a/test/widgets/main_test.dart
+++ b/test/widgets/main_test.dart
@@ -5,6 +5,7 @@ import 'package:motis_mitfahr_app/main_app.dart';
 import 'package:motis_mitfahr_app/util/locale_manager.dart';
 import 'package:motis_mitfahr_app/util/supabase.dart';
 import 'package:motis_mitfahr_app/util/theme_manager.dart';
+import 'package:motis_mitfahr_app/welcome/pages/reset_password_page.dart';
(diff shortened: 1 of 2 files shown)


Eddie-42 (approved):

Showcas (approved):

test/widgets/main_test.dart
```diff
@@ -0,0 +1,179 @@
...
+      localeManager.setCurrentLocale(const Locale('de'));
+      // ignore: invalid_use_of_protected_member
+      expect(themeManager.hasListeners, isTrue);
+      themeManager.setTheme(ThemeMode.dark);
+
+      final Finder materialApp = find.byType(MaterialApp);
+      expect(materialApp, findsOneWidget);
+      final MaterialApp app = materialApp.evaluate().first.widget as MaterialApp;
+      expect(app.home.runtimeType, AuthApp);
+      expect(app.theme, themeManager.lightTheme);
```
istzen, line 41: why is this ligthTheme? i thought you set the theme to dark in line 35.

fynsta: "theme" just means "lightTheme". "themeMode" would be "currentTheme".
No idea why the naming is so bad. I changed this line to check the dark theme to be a bit more explicit.

I would have loved to use "Theme.of(context)" to get the current theme of the app, but that did not work for some reason.

test/widgets/main_test.dart
```diff
@@ -0,0 +1,179 @@
...
+      expect(resetPasswordPage, findsOneWidget);
+      final ResetPasswordPage page = resetPasswordPage.evaluate().first.widget as ResetPasswordPage;
+      page.onPasswordReset();
+
+      await tester.pump();
+
+      expect(find.byType(MainApp), findsOneWidget);
+    });
+
+    // TODO: Write test case that checks the snackbar when email is invalid/expired
```
fynsta, line 176: Hey @Eddie-42, I just tried to test this case.

The thing is: I don't think the case is reachable? I've looked into the code and it definitely seems impossible. I looked it up and saw that you added it with this commit: https://github.com/MOTIS-Mitfahr-App/flutter-app/commit/f7abadf01bf05991a633ce2844226a580c679678

Do you remember this? Did you try this out back then?

Eddie-42: I have no strong recollection, but I'm pretty sure I never tested it out. I think this might be copied from a site in the supabase docs or something. But why is it impossible to reach?

fynsta: I can show you in person, it's kind of difficult to explain on Github

MarkFalke (approved):

test/widgets/main_test.dart
fynsta: I just tested it, I could not get this case. So I'm just going to remove it. (And I could not find the site in the docs where you could have copied this from).

PR #96, opened 2023-01-28
Unify snackbars

fynsta:
This unifies the snackbar behaviour. It will always announce the message via the SemanticsService (are there cases where we do not want that @Eddie-42 ?). There are also 3 predefined lengths of visibility for the SnackBar.

Also, adds a "replace" parameter that allows to replace the whole SnackBar queue (and hiding the current SnackBar). That's especially important in the "EditProfileFeatures" page where it was previously possible to create an insanely long queue.

### DIFF
--- a/lib/drives/pages/create_drive_page.dart
+++ b/lib/drives/pages/create_drive_page.dart
@@ -9,6 +9,7 @@ import '../../util/fields/increment_field.dart';
 import '../../util/locale_manager.dart';
 import '../../util/search/address_suggestion.dart';
 import '../../util/search/start_destination_timeline.dart';
+import '../../util/snackbar.dart';
 import '../../util/supabase.dart';
 import '../../util/trip/trip.dart';
 import '../models/drive.dart';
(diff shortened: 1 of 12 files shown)


lib/drives/pages/create_drive_page.dart
```diff
@@ -170,10 +171,9 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
         );
       } on AuthException {
         //todo: change error message when login is implemented
-        ScaffoldMessenger.of(context).showSnackBar(
```
istzen, line 173: looking over the code i'm not sure if we even need this part. What could be an Exception here? I mean that all values are valid is checked by the validators and. If I have the rights to create a drive is checked by the RLS and that should not return an Exception.

istzen (commented):
I like this. We should maybe also do this for other stuff, like Padding, font size etc.

Eddie-42 (approved):

MarkFalke (approved):

fynsta:
> I like this. We should maybe also do this for other stuff, like Padding, font size etc.

Yes but those should be defined in a constants file somewhere, that's a little bit different.

lib/drives/pages/create_drive_page.dart
fynsta: Do you think so? Now I'm not sure, what is returned on `insert` when RLS prevents it?

But that should be done in another PR anyways.

PR #100, opened 2023-01-31
Better PasswordField (User study feedback)

fynsta:
- Main thing: Add a toggle button for obscuring the password. This also required the widget to become stateful, of course.
- Refactor the PasswordField because the validator logic was always one of three things. Now the pages just need to say whether they want to validate strictly (Reset + Register vs. Login) and if the password is a confirmation field (by providing the controller of the original password field)
- Adds a HelperText for the strictly validating fields to make the requirements more clear. It actually updates on change and always tells you the next requirement to fulfill.
- Refactor the tests into their own file because there was a lot of duplication (only possible now that the logic actually resides inside the widget but I remember @istzen asking for it on the original PR)

I also saw that it is possible to do form validation on user input, is that something we should maybe do?

### DIFF
--- a/lib/util/fields/password_field.dart
+++ b/lib/util/fields/password_field.dart
@@ -101,7 +101,7 @@ class _PasswordFieldState extends State<PasswordField> {
     } else if (value.length < 8) {
       return S.of(context).formPasswordValidateMinLength;
     } else if (RegExp('[0-9]').hasMatch(value) == false) {
-      return S.of(context).formPasswordValidateMinLength;
+      return S.of(context).formPasswordValidateNumber;
     } else if (RegExp('[A-Z]').hasMatch(value) == false) {
       return S.of(context).formPasswordValidateUppercase;
     } else if (RegExp('[a-z]').hasMatch(value) == false) {
--- a/lib/util/fields/password_field.dart
+++ b/lib/util/fields/password_field.dart
@@ -5,15 +5,15 @@ class PasswordField extends StatefulWidget {
   static const Key passwordFieldKey = Key('passwordField');
   static const Key passwordConfirmationFieldKey = Key('passwordConfirmationField');
 
-  final bool validateStrictly;
-  final TextEditingController? controller;
+  final TextEditingController controller;
   final TextEditingController? originalPasswordController;
+  final bool validateStrictly;
 
   const PasswordField({
     super.key,
-    this.validateStrictly = false,
+    required this.controller,
     this.originalPasswordController,
-    this.controller,
+    this.validateStrictly = false,
   }) : assert(validateStrictly == false || originalPasswordController == null);
 
   @override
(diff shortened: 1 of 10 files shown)


lib/util/fields/password_field.dart
```diff
@@ -1,39 +1,114 @@
...
+    return null;
+  }
+
+  String? _validateStrictly(String? value) {
+    if (value == null || value.isEmpty) {
+      return S.of(context).formPasswordValidateEmpty;
+    } else if (value.length < 8) {
+      return S.of(context).formPasswordValidateMinLength;
+    } else if (RegExp('[0-9]').hasMatch(value) == false) {
+      return S.of(context).formPasswordValidateMinLength;
```
Showcas, line 104: This should not be the right Return is it? 
use formPasswordValidateNumber

lib/util/fields/password_field.dart
```diff
@@ -1,39 +1,114 @@
...
+  }
+
+  String? _validateOnlyEmpty(String? value) {
+    if (value == null || value.isEmpty) {
+      return S.of(context).formPasswordValidateEmpty;
+    }
+    return null;
+  }
+
+  String? _validateStrictly(String? value) {
```
Showcas, line 98: A general question: an other User Study gave feedback that they prefer to know with the first error what all the requirement for the password are. Should we include this in this PR or do you prefer to leave it as it is?

lib/util/fields/password_field.dart
```diff
@@ -1,39 +1,114 @@
...
-  final String? errorText;
-  final String? helperText;
-  final String? Function(String?)? validator;
-  final TextEditingController? controller;
+class PasswordField extends StatefulWidget {
+  static const Key passwordFieldKey = Key('passwordField');
+  static const Key passwordConfirmationFieldKey = Key('passwordConfirmationField');
+
+  final TextEditingController controller;
+  final TextEditingController? originalPasswordController;
```
Eddie-42, line 9: I don't think this is a good name. I don't have any GREAT ideas for alternatives, but this sounds to me like there is an "original" and an "actual" password controller for the same field

lib/util/fields/password_field.dart
```diff
@@ -1,39 +1,114 @@
...
+  final bool validateStrictly;
```
Eddie-42, line 10: This, too. Maybe `validateSecurity`?

lib/util/fields/password_field.dart
Eddie-42: We do have a version of this now don't we? At least you don't have to submit the form to know that the password isn't secure enough yet. I think the current implementation is very cool. Communicating every requirement right away is a bad idea IMO because 1) It overloads the user with information they don't need yet and 2) There is no way to display that really in an acceptable way

Eddie-42 (approved):

lib/util/fields/password_field.dart
fynsta: I thought about `passwordForConfirmationController` which would be accurate, but very confusing because it sounds like the exact opposite of what it does. I can definitely add a comment here, though.

lib/util/fields/password_field.dart
fynsta: Lol yes great catch. I had copied that from our existing pages, where it was wrong as well. How did no one ever notice that??

fynsta: That was a nifty copy-paste error from back when I added the translations. Glad you noticed!

istzen (approved):

Showcas (approved):

PR #102, opened 2023-01-31
Cleans up remaining TODOS

fynsta:
- AG did not want TOS link
- Added correct "Withdraw button" text in FAQ
- added CopyWith for Ride and status/id, if needed please add for other models and attributes but that code is just not worth it right now
- removed unnecessary escapes in email regex

Removed all other todos because they were a little bit random. Of course we still need to add the algorithm, but we do not need to write that in the code. Same with notifications.

### DIFF
--- a/lib/rides/models/ride.dart
+++ b/lib/rides/models/ride.dart
@@ -173,6 +173,29 @@ class Ride extends Trip {
   String toString() {
     return 'Ride{id: $id, in: $driveId, from: $start at $startTime, to: $end at $endTime, by: $riderId}';
   }
+
+  Ride copyWith({int? id, RideStatus? status}) {
+    return Ride(
+      id: id ?? this.id,
+      createdAt: createdAt,
+      start: start,
+      startPosition: startPosition,
+      startTime: startTime,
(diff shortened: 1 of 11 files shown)


Showcas (approved):

Eddie-42 (approved):

lib/rides/models/ride.dart
```diff
@@ -173,6 +173,29 @@ class Ride extends Trip {
   String toString() {
     return 'Ride{id: $id, in: $driveId, from: $start at $startTime, to: $end at $endTime, by: $riderId}';
   }
+
+  Ride copyWith({int? id, RideStatus? status}) {
```
istzen, line 177: should we add a copyWith for every Model?

fynsta: Theoretically, yes. But as long as it's not needed, I would not do it (see PR description). Same with the attributes that can be changed. I only did those that are currently needed, furthers should be added on demand.

MarkFalke (approved):

fynsta:
@istzen everything fine?

PR #103, opened 2023-01-31
Enable remaining analysis rules

fynsta:
- `avoid_classes_with_only_static_members`: I converted StorageManager & SupabaseManager to default instance exports
- `use_setters_to_change_properties`. Just changed the assignment for SupabaseManager setClient and setCurrentProfile. Now please use `supabaseManager.currentProfile = ... ` and `supabaseManager.client = ... `. Same for the getters, of course.

I also had to move some code into the initialize method to make sure that "Supabase.instance.client" is not called inside the tests, where it would not have been initialized.

Finally, I renamed the file to include the "Manager" name

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -12,7 +12,7 @@ linter:
     - always_specify_types
     - annotate_overrides
     - avoid_bool_literals_in_conditional_expressions
-    # - avoid_classes_with_only_static_members # ENABLE IN THE FUTURE
+    - avoid_classes_with_only_static_members
     - avoid_double_and_int_checks
     - avoid_dynamic_calls
     - avoid_empty_else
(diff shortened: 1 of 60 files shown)


Eddie-42 (approved):

istzen (approved):
Just for me to understand this. why is it better to have a supabaseManager instance and no get and set method for current profile?

fynsta:
https://dart-lang.github.io/linter/lints/avoid_classes_with_only_static_members.html
> Creating classes with the sole purpose of providing utility or otherwise static methods is discouraged. Dart allows functions to exist outside of classes for this very reason.

And these get/set methods are basically just unnecessary because Dart already supplies getters/setters for public attributes via the standard syntax

MarkFalke (approved):

PR #106, opened 2023-01-31
Fix order of waypoints in DriveDetailPage

fynsta:
Before, it could happen that the end of the whole drive was placed before another Waypoint (of course only if the time is the same, which should probably not happen in reality - but it happened all the time with our fake data).


fynsta:
I also added a test case that was failing with the previous code. Now the test actually tests which WaypointCards are built and in which order.

Showcas (approved):

istzen (approved):

PR #107, opened 2023-02-01
Generate trip times better

fynsta:
As talked about. If startTime is given, then endTime is generated from that. If endTime is given, then startTime is generated from that.

Also adds a convenience "duration" parameter, let's see if that one is useful.

### DIFF
--- a/test/util/factories/drive_factory.dart
+++ b/test/util/factories/drive_factory.dart
@@ -27,8 +27,6 @@ class DriveFactory extends TripFactory<Drive> {
     NullableParameter<Profile>? driver,
     List<Ride>? rides,
     List<Chat>? chats,
-    // Convenience parameter to set the duration of the trip
-    Duration? duration,
     bool createDependencies = true,
   }) {
     assert(driverId == null || driver?.value == null || driver!.value?.id == driverId);
@@ -47,7 +45,7 @@ class DriveFactory extends TripFactory<Drive> {
               )
             : null);
 
-    final TripTimes tripTimes = generateTimes(startTime, endTime, duration);
+    final TripTimes tripTimes = generateTimes(startTime, endTime);
 
     return Drive(
       id: id ?? randomId,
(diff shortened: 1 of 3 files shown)


Showcas (approved):

test/util/factories/drive_factory.dart
```diff
@@ -27,6 +27,8 @@ class DriveFactory extends TripFactory<Drive> {
     NullableParameter<Profile>? driver,
     List<Ride>? rides,
     List<Chat>? chats,
+    // Convenience parameter to set the duration of the trip
+    Duration? duration,
```
Eddie-42, line 31: Oof this hurts a little. Every other parameter is an actual parameter of the class we are generating or createDependencies, and in comes this exception 😬

fynsta: I mean we can talk about creating a `generateFakefromDuration` method, but come on - then we would have to write down every attribute _again_

Eddie-42: > Yes but that's why I added the comment. I was not 100% sure, but it might be very useful for readability of the test code.
> 
> (Also, not entirely true: See `createDependencies`)

Haha I mentioned `createDependencies`

fynsta: Oh I can't read obviously... Probably only because you did not write it in code style :|

Eddie-42: > I mean we can talk about creating a `generateFakefromDuration` method, but come on - then we would have to write down every attribute _again_

That would also hurt a little. I don't know what you'd expect here if you were an experienced test writer...

istzen: i am not sure why we need to set the duration at the moment, but i guess it's okay like this with the comment

istzen (approved):

test/util/factories/drive_factory.dart
fynsta: I'll remove it, and we can add it again when we think of a good usecase

Eddie-42 (approved):

PR #112, opened 2023-02-02
Automatic overall review if category reviews given

Eddie-42:
Suggestion came from a user study

### DIFF
--- a/lib/account/pages/write_review_page.dart
+++ b/lib/account/pages/write_review_page.dart
@@ -21,6 +21,7 @@ class WriteReviewPage extends StatefulWidget {
 }
 
 class _WriteReviewPageState extends State<WriteReviewPage> {
+  bool reviewChanged = false;
   Review? _review;
   late TextEditingController _textController;
   ButtonState _state = ButtonState.idle;
@@ -148,32 +149,49 @@ class _WriteReviewPageState extends State<WriteReviewPage> {
   }
 
   void onRatingUpdate(double rating) {
+    reviewChanged = true;
     setState(() {
       _review!.rating = rating.toInt();
     });
   }
 
+  void _updateOverallReview() {
+    if (!reviewChanged) {
+      _review!.rating = ((_review!.comfortRating ?? 0) +
+              (_review!.safetyRating ?? 0) +
+              (_review!.hospitalityRating ?? 0) +
+              (_review!.reliabilityRating ?? 0)) ~/
+          <int?>[_review!.comfortRating, _review!.safetyRating, _review!.hospitalityRating, _review!.reliabilityRating]
+              .where((int? element) => element != null)
+              .length;
+    }
+  }
+
   void onComfortRatingUpdate(double rating) {
     setState(() {
       _review!.comfortRating = rating.toInt();
+      _updateOverallReview();
     });
   }
 
   void onSafetyRatingUpdate(double rating) {
     setState(() {
       _review!.safetyRating = rating.toInt();
+      _updateOverallReview();
     });
   }
 
   void onReliabilityRatingUpdate(double rating) {
     setState(() {
       _review!.reliabilityRating = rating.toInt();
+      _updateOverallReview();
     });
   }
 
   void onHospitalityRatingUpdate(double rating) {
     setState(() {
       _review!.hospitalityRating = rating.toInt();
+      _updateOverallReview();
     });
   }
 
--- a/lib/account/pages/write_review_page.dart
+++ b/lib/account/pages/write_review_page.dart
@@ -164,7 +164,8 @@ class _WriteReviewPageState extends State<WriteReviewPage> {
         _review!.reliabilityRating
       ].where((int? element) => element != null);
       _review!.rating =
-          givenCategoryRatings.reduce((int? value, int? element) => value! + element!)! ~/ givenCategoryRatings.length;
+          (givenCategoryRatings.reduce((int? value, int? element) => value! + element!)! / givenCategoryRatings.length)
+              .round();
     }
   }
 
--- a/lib/account/pages/write_review_page.dart
+++ b/lib/account/pages/write_review_page.dart
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:progress_state_button/progress_button.dart';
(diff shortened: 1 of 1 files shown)


lib/account/pages/write_review_page.dart
```diff
@@ -21,6 +21,7 @@ class WriteReviewPage extends StatefulWidget {
 }
 
 class _WriteReviewPageState extends State<WriteReviewPage> {
+  bool reviewChanged = false;
```
fynsta, line 24: `reviewChangedManually`. It changes all the time

lib/account/pages/write_review_page.dart
```diff
@@ -148,32 +149,49 @@ class _WriteReviewPageState extends State<WriteReviewPage> {
...
   void onRatingUpdate(double rating) {
+    reviewChanged = true;
     setState(() {
       _review!.rating = rating.toInt();
     });
   }
 
+  void _updateOverallReview() {
+    if (!reviewChanged) {
+      _review!.rating = ((_review!.comfortRating ?? 0) +
```
fynsta, line 160: I think this code could be improved in readability.
Something like
```
_review!.rating = <int?>[_review!.comfortRating, _review!.safetyRating, _review!.hospitalityRating,
_review!.reliabilityRating].where((int? element) => element != null).average;
```

lib/account/pages/write_review_page.dart
fynsta: also, `rating` instead of `review`

fynsta:
Actually I'm not sure I like the whole concept. It would make more sense if this was only applied when you send the review. Otherwise these values are just jumping around.

Eddie-42:
> Actually I'm not sure I like the whole concept. It would make more sense if this was only applied when you send the review. Otherwise these values are just jumping around.

I don't get the problem. I think your suggestion would be worse in every way: It would essentially do the same, but provide no feedback to the user whatsoever. I think the user deserves to know that what we're actually doing is automatically setting the overall rating, not leaving it empty. I imagine the current behavior is useful for two usecases: 1) First-time users, where this essentially says "Hey, it's great that you're setting the category ratings, but you've got to set the overall rating as well. We've provided you a suggestion!", and 2) Experienced users who might appreciate having to tap one less time if they're setting every rating every time. For 1), your solution would be worse because they're not getting the feedback right away, and for 2) it's worse because now they have to tap one more time again

lib/account/pages/write_review_page.dart
```diff
@@ -148,32 +149,50 @@ class _WriteReviewPageState extends State<WriteReviewPage> {
...
+  void _updateOverallReview() {
+    if (!ratingChangedManually) {
+      final Iterable<int?> givenCategoryRatings = <int?>[
+        _review!.comfortRating,
+        _review!.safetyRating,
+        _review!.hospitalityRating,
+        _review!.reliabilityRating
+      ].where((int? element) => element != null);
+      _review!.rating =
+          givenCategoryRatings.reduce((int? value, int? element) => value! + element!)! ~/ givenCategoryRatings.length;
```
istzen, line 167: this is always rounding down, is this what we want, i think it' confusing if i select 3 stars everywhere except one with 2 and the overall rating is 2

fynsta: Why are we rounding at all?

fynsta:
> I don't get the problem. I think your suggestion would be worse in every way: It would essentially do the same, but provide no immediate feedback to the user. I imagine the current behavior is useful for two usecases: 1) First-time users, where this essentially says "Hey, it's great that you're setting the category ratings, but you've got to set the overall rating as well. We've provided you a suggestion!", and 2) Experienced users who might appreciate having to tap one less time if they're setting every rating every time. For 1), your solution would be worse because they're not getting the feedback right away, and for 2) it's worse because now they have to tap one more time again

Okay but I'm still not entirely convinced. Maybe we should show the automatically calculated rating in a slightly different color/style?

lib/account/pages/write_review_page.dart
Eddie-42: Because the user can't input double values, it'd be weird if they were only achievable this way?

lib/account/pages/write_review_page.dart
fynsta: Btw I wasn't kidding with average (part of the collection package). Here is working code:

```
      
final List<int> givenCategoryRatings = <int?>[
        _review!.comfortRating,
        _review!.safetyRating,
        _review!.hospitalityRating,
        _review!.reliabilityRating
      ].whereType<int>().toList();
_review!.rating = givenCategoryRatings.average.round();
```

Eddie-42:
> > I don't get the problem. I think your suggestion would be worse in every way: It would essentially do the same, but provide no immediate feedback to the user. I imagine the current behavior is useful for two usecases: 1) First-time users, where this essentially says "Hey, it's great that you're setting the category ratings, but you've got to set the overall rating as well. We've provided you a suggestion!", and 2) Experienced users who might appreciate having to tap one less time if they're setting every rating every time. For 1), your solution would be worse because they're not getting the feedback right away, and for 2) it's worse because now they have to tap one more time again
> 
> Okay but I'm still not entirely convinced. Maybe we should show the automatically calculated rating in a slightly different color/style?

I still don't see the problem! I think a different color/style is weird because you might think it will somehow be stored differently/you have to input it because this is just a suggestion. What's wrong with the current state?

lib/account/pages/write_review_page.dart
fynsta: I think it would be fun. The average might be a whole half star away from what we will show. That's not particularly accurate...

fynsta:
> color/style

Currently it feels really weird that user input is not distinguishable from automatic input. And I actually think what you are saying is exactly what I'm trying to do: Make it more obvious that this is a suggestion, and that the user can change it at will.

Eddie-42:
> Currently it feels really weird that user input is not distinguishable from automatic input. And I actually think what you are saying is exactly what I'm trying to do: Make it more obvious that this is a suggestion, and that the user can change it at will.

Yes but a rating bar that the user can change at will is exactly the style we're using right now for every other rating bar that the user can change at will, like for the categories. I don't mind that the user input isn't distinguishable from the automatic input and I think it is actually less intuitive if people have to think about why there are two different kinds of input presented to them. They might even come to the conclusion that whatever we are using for the automatic rating is just the normal input for the overall rating. Also, the automatic input essentially represents the user input, and I think there are more use cases where you'd want to change our automatic suggestion than just leave it. So I don't want to stress too much that you can change it.

istzen (approved):

lib/account/pages/write_review_page.dart
istzen: yes that's true but since the user can't input double values it makes more sense this way in my opinion.

fynsta (approved):

fynsta:
Hm okay I'm still not 100% sure this is an improvement, but fine

Showcas (approved):

PR #113, opened 2023-02-02
Flutter 3.7

fynsta:
The update forced a few differences:

 - Different textTheme names (previously this was super confusing, but I think I used the right ones now)
 - errorColor is deprecated, we should colorScheme.error (we wanted that anyways)
 - DIfferent arb format ("one" instead of "=1", no comma between the different cases, no placeholders needed anymore)
 - Flutter has a "Badge" class itself now. I tried it out and it did not work particularly well, so I would stay with our package for now. But the import needs to be named for this to work from now on.

Please check if everything is alright. The app style looks a little bit different, not sure how much that should actually change.


istzen:
i'm not sure about the primary color i think, it looks very similar to the error color. Especially in the dark mode.

Showcas (approved):
looks good to me, tested most features. We maybe want to talk about the colors in a meeting but the rest looks good

PR #114, opened 2023-02-02
Test search ride

Eddie-42:
Search Ride and Search Ride Filter are 100% tested*, and I kept the test file just under 1000 lines!

* I found one (VERY minor) with the spacing in the indicator row.
* I refactored the swap start and destination button into the StartDestinationTimeline.
* The price is now actually dependent on the distance. I needed that in order to test the ordering by price. Because we want to test every function we add from now on, I also tested the Position model.
* I added a goodie: If you look for a ride and there is one present but not at the time you selected, it will tell you to try another time, now (instead of the "Relax your filters" text, which I noticed while testing wasn't really fitting). If you tap on the text, it will automatically take you to the next-best time.

\* If you test this between 00:00 and 00:10, there is one line that isn't tested. The line in question is the validator telling you to choose a valid time, as you chose a time more than 10 minutes ago. Because we use DateTime.now() in that test and I didn't find a good way to mock that online, I thought I'd leave it as is for now. The internet recommends using dart.clock if you want to mock the time afaik.

### DIFF
--- a/lib/rides/pages/search_ride_page.dart
+++ b/lib/rides/pages/search_ride_page.dart
@@ -33,16 +33,16 @@ class SearchRidePageState extends State<SearchRidePage> {
   bool _wholeDay = true;
   int seats = 1;
 
-  late final SearchRideFilter _filter;
+  late final SearchRideFilter filter;
 
-  List<Ride>? _rideSuggestions;
+  List<Ride>? rideSuggestions;
 
   bool _loading = false;
 
   @override
   void initState() {
     super.initState();
-    _filter = SearchRideFilter(wholeDay: _wholeDay);
+    filter = SearchRideFilter(wholeDay: _wholeDay);
     loadRides();
   }
 
@@ -127,12 +127,12 @@ class SearchRidePageState extends State<SearchRidePage> {
             drive.endTime,
             seats,
             supabaseManager.currentProfile?.id ?? -1,
-            10.25,
+            double.parse(drive.startPosition.distanceTo(drive.endPosition).toStringAsFixed(2)),
           ),
         )
         .toList();
     setState(() {
-      _rideSuggestions = rides;
+      rideSuggestions = rides;
       _loading = false;
     });
   }
@@ -234,7 +234,7 @@ class SearchRidePageState extends State<SearchRidePage> {
             value: _wholeDay,
             onChanged: (bool? value) => setState(() {
               _wholeDay = value!;
-              _filter.wholeDay = _wholeDay;
+              filter.wholeDay = _wholeDay;
             }),
           ),
         ],
@@ -284,7 +284,7 @@ class SearchRidePageState extends State<SearchRidePage> {
   }
 
   Widget buildMainContentSliver() {
-    if (_rideSuggestions == null || _loading) {
+    if (rideSuggestions == null || _loading) {
       if (_startSuggestion == null || _destinationSuggestion == null) {
         return buildEmptyRideSuggestions(
           key: const Key('searchRideNoInput'),
@@ -301,7 +301,7 @@ class SearchRidePageState extends State<SearchRidePage> {
         ),
       );
     }
-    final List<Ride> filterApplied = _filter.apply(_rideSuggestions!, selectedDate);
+    final List<Ride> filterApplied = filter.apply(rideSuggestions!, selectedDate);
     final List<Ride> filteredSuggestions = applyTimeConstraints(filterApplied);
     if (filteredSuggestions.isEmpty) {
       return buildEmptyRideSuggestions(
@@ -329,13 +329,13 @@ class SearchRidePageState extends State<SearchRidePage> {
                   ),
                 ),
               )
-            : _rideSuggestions!.isNotEmpty
+            : rideSuggestions!.isNotEmpty
                 ? Semantics(
                     button: true,
                     tooltip: S.of(context).pageSearchRideTooltipFilter,
                     child: InkWell(
                       key: const Key('searchRideRelaxRestrictions'),
-                      onTap: () => _filter.dialog(context, setState),
+                      onTap: () => filter.dialog(context, setState),
                       child: Text(
                         S.of(context).pageSearchRideRelaxRestrictions,
                         style: Theme.of(context).textTheme.titleMedium,
@@ -391,7 +391,7 @@ class SearchRidePageState extends State<SearchRidePage> {
               SliverToBoxAdapter(
                 child: Padding(
                   padding: const EdgeInsets.symmetric(horizontal: 8),
-                  child: _filter.buildIndicatorRow(context, setState),
+                  child: filter.buildIndicatorRow(context, setState),
                 ),
               ),
               SliverPinnedHeader(
--- a/lib/rides/pages/search_ride_page.dart
+++ b/lib/rides/pages/search_ride_page.dart
@@ -23,9 +23,9 @@ class SearchRidePage extends StatefulWidget {
 
 class SearchRidePageState extends State<SearchRidePage> {
   final TextEditingController startController = TextEditingController();
-  AddressSuggestion? startSuggestion;
+  AddressSuggestion? _startSuggestion;
   final TextEditingController destinationController = TextEditingController();
-  AddressSuggestion? destinationSuggestion;
+  AddressSuggestion? _destinationSuggestion;
 
   final TextEditingController _dateController = TextEditingController();
   final TextEditingController _timeController = TextEditingController();
@@ -103,7 +103,7 @@ class SearchRidePageState extends State<SearchRidePage> {
   }
 
   Future<void> loadRides() async {
-    if (startSuggestion == null || destinationSuggestion == null) return;
+    if (_startSuggestion == null || _destinationSuggestion == null) return;
     setState(() => _loading = true);
     final List<Map<String, dynamic>> data =
         await supabaseManager.supabaseClient.from('drives').select<List<Map<String, dynamic>>>('''
@@ -119,11 +119,11 @@ class SearchRidePageState extends State<SearchRidePage> {
         .map(
           (Drive drive) => Ride.previewFromDrive(
             drive,
-            startSuggestion!.name,
-            startSuggestion!.position,
+            _startSuggestion!.name,
+            _startSuggestion!.position,
             drive.startTime,
-            destinationSuggestion!.name,
-            destinationSuggestion!.position,
+            _destinationSuggestion!.name,
+            _destinationSuggestion!.position,
             drive.endTime,
             seats,
             supabaseManager.currentProfile?.id ?? -1,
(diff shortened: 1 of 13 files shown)


istzen:
the searchFillter sort price test has a warning. Can you remove that?

lib/rides/widgets/search_ride_filter.dart
```diff
@@ -300,9 +314,7 @@ class SearchRideFilter {
         if (_minComfortRating != _defaultRating ||
             _minSafetyRating != _defaultRating ||
             _minReliabilityRating != _defaultRating ||
-            _minHospitalityRating != _defaultRating) {
-          ratingWidgets.add(const SizedBox(width: 4));
```
fynsta, line 304: this was removed for what reason? Now the `if` is useless as far as i can see

test/widgets/search_ride_page_test.dart
```diff
@@ -0,0 +1,967 @@
...
+            expect(pageState.selectedDate.month, dateTime.month);
+            expect(pageState.selectedDate.day, dateTime.day);
+            expect(pageState.selectedDate.hour, dateTime.hour);
+            expect(pageState.selectedDate.minute, dateTime.minute);
+          });
+
+          testWidgets('Impossible time', (WidgetTester tester) async {
+            final DateTime dateTime = DateTime.now().subtract(const Duration(minutes: 11));
+            if (dateTime.day != DateTime.now().day) {
+              // Code is untestable from 00:00 to 00:11
```
fynsta, line 306: I know we talked about this... But I am unhappy that the line is just not part of the coverage when we run it at that time - I feel like coverage should be predetermined

test/widgets/search_ride_page_test.dart
```diff
@@ -0,0 +1,967 @@
...
+        await enterFilter(tester, rating: 3);
+        expect(find.descendant(of: indicatorRow, matching: find.byIcon(Icons.star)), findsOneWidget);
+
+        await enterFilter(tester, rating: 3, comfortRating: 3, hospitalityRating: 3);
+        expect(find.descendant(of: indicatorRow, matching: find.byIcon(Icons.star)), findsNWidgets(3));
+        expect(find.descendant(of: indicatorRow, matching: find.byIcon(Icons.chair)), findsOneWidget);
+        expect(find.descendant(of: indicatorRow, matching: find.byIcon(Icons.favorite)), findsOneWidget);
+        for (int i = 0; i < 2; i++) {
+          expect(find.descendant(of: indicatorRow, matching: find.byKey(Key('ratingSizedBox$i'))), findsOneWidget);
+        }
```
fynsta, line 503: lol this is longer than the two liens would have been

test/widgets/search_ride_page_test.dart
```diff
@@ -0,0 +1,967 @@
...
+          await pumpMaterial(tester, const SearchRidePage());
+
+          await tester.tap(find.byKey(const Key('searchRideFilterButton')));
+          await tester.pump();
+
+          final Finder featuresExpandButton = find.byKey(const Key('searchRideFeaturesExpandButton'));
+          await tester.tap(featuresExpandButton);
+          await tester.pump();
+
+          final Finder featureChip = find.byKey(Key('searchRideFeatureChip${Feature.values[0].name}'));
```
fynsta, line 563: should probably move `Feature.values[0]` to a variable for this test group.

test/widgets/search_ride_page_test.dart
```diff
@@ -0,0 +1,967 @@
...
+              null) {
+            mutuallyExclusiveFiltered.add(feature);
+          }
+        }
+        final List<Feature> requiredFeatures =
+            mutuallyExclusiveFiltered.sublist(random.integer(mutuallyExclusiveFiltered.length - 1));
+
+        final List<Profile> drivers = [
+          //Satisfies everything
+          ProfileFactory().generateFake(
```
fynsta, line 619: Just define these with names that say what they are doing, then you can remove the comments. (and the code with `drivers[3]` is more readable, for example)

test/widgets/search_ride_page_test.dart
```diff
@@ -0,0 +1,967 @@
...
+
+        await pumpMaterial(tester, const SearchRidePage());
+
+        await enterStartAndDestination(tester, faker.address.city(), faker.address.city());
+        await enterFilter(tester, features: [Feature.values[random.integer(Feature.values.length)]]);
+
+        expect(find.byType(RideCard, skipOffstage: false), findsNothing);
+
+        final Finder relaxRestrictionsButton = find.byKey(const Key('searchRideRelaxRestrictions')).hitTestable();
+        await tester.scrollUntilVisible(relaxRestrictionsButton, 100, scrollable: find.byType(Scrollable).first);
```
fynsta, line 440: It should be noted that this is needed because the test screen is so small - on a normal screen the button would be visible, but here we have to scroll to it (@everyone)

test/widgets/search_ride_page_test.dart
Eddie-42: Yes I'm not happy with it either, but I didn't want to implement anything before we talked about it because I think the solution isn't beautiful and it's important for further development. As discussed [here](https://stackoverflow.com/questions/53816067/how-to-test-code-that-uses-datetime-now-in-flutter), we could
1. use the dart `clock` class instead of `DateTime`. I don't like that because I feel it's not very Flutter-y, everywhere I look everyone uses `DateTime`.
2. implement an extension method on `DateTime` and use that. I don't like that because I don't think there's a good way to make the difference between a mockable now and an un-mockable now very clear, and people will be wondering why we used a strange method at first sight. Besides, we'd essentially be re-implementing `clock`, I imagine.
3. Test if the program is currently in test mode via the environment variable and use either `DateTime.now` or e.g. clock
4. Ignore the problem (our actual code will be slightly more streamlined in any case IMO) and accept the fact that coverage will be 0.0x% worse between 0:00 and 0:11.

Either way, I'll maybe have to change every usage of `DateTime.now` for consistency reasons, and I felt like that would have been too big a change not to discuss.

test/widgets/search_ride_page_test.dart
Eddie-42: Yes but I think it's okay. I think it's in line with some other test I copied it from

test/widgets/search_ride_page_test.dart
Eddie-42: Yes, that sent me into a spin for quite some time because it appears the screen size is different if you execute tests one by one vs. if you execute the whole group because when I executed the test on its own, it worked fine every time, but within the group, it failed every time. So if you're experiencing anything like that, maybe that's your solution

lib/rides/widgets/search_ride_filter.dart
Eddie-42: Haha! Yeah you're right and I removed the `if`. That was the tiny spacing issue I was talking about. Turns out, I counted the length of the `ratingWidgets` in order to loop over it when inserting `SizedBox`es between every category review. This line was supposed to add a little more spacing between the overall and category ratings. But turns out, it adds a ratingWidget I failed to account for and there was always one `SizedBox` too much at the end. Anyways, I realized how dumb that was and now I'm adding the extra "overallRating" `SizedBox` _after_ the other `SizedBox`es

test/widgets/search_ride_page_test.dart
```diff
@@ -0,0 +1,972 @@
...
+          rating: minRating,
+          comfortRating: minRating,
+          safetyRating: minRating,
+          reliabilityRating: minRating,
+          hospitalityRating: minRating,
+        );
+
+        List<Ride> filteredRides = pageState.filter.apply(pageState.rideSuggestions!, pageState.selectedDate);
+        expect(filteredRides, hasLength(3));
+        expect(filteredRides[0].driveId, drives[0].id);
```
fynsta, line 721: Still not happy you need to use indices here instead of the name, but I guess there is no way to get to the drive of that driver?

fynsta: Maybe use `Map.fromIterables`?

Eddie-42:
I tried around a little bit and can't seem to figure out what causes the behavior the warning describes. Unfortunately, I am unable to work on this much in the coming days, so I'm considering skipping the test and dealing with it later

fynsta (approved):

Showcas (approved):

test/models/position_test.dart
```diff
@@ -0,0 +1,67 @@
...
+      final Position position1 = Position(50, 8.2714);
+      final Position position2 = Position(50, 8.2715);
+      expect(position1 == position2, isFalse);
+    });
+  });
+
+  group('Position.hashCode', () {
+    test('Mainz', () {
+      final Position position = Position(50, 8.2714);
+      expect(position.hashCode, 9159982799919094);
```
Eddie-42, line 55: I had to change this value just now because I think the new dart version just returned something different. That reminds me that I actually hate this test, does anyone have a better idea how to test this or why we even have the method `Position.hashCode`?

fynsta: You need to implement `hashCode` when you implement the `==` operator. Maybe that stupid fact alone would be reason to use this package instead: https://pub.dev/packages/equatable (Which generates those for you)

fynsta: I think you just need to make sure that hashCode is equal if `==` considers the both equal (keep in mind, for some reason not necessarily _exactly if_, just if). So maybe just test that fact?

Eddie-42: So
```
final Position position1 = Position(50, 8.2714);
final Position position2 = Position(50, 8.2714);
expect(position1.hashCode, position2.hashCode);
```
? Boy do I not like this test

fynsta: Yes that seems like a better test. And then

```
final Position otherPosition = Position(50, 8.2715);
expect(position1.hashCode == position2.hashCode, false);
```

lib/rides/pages/search_ride_page.dart
```diff
@@ -15,7 +16,9 @@ import '../models/ride.dart';
 import '../widgets/search_ride_filter.dart';
 
 class SearchRidePage extends StatefulWidget {
-  const SearchRidePage({super.key});
+  //This is needed in order to mock the time in tests
+  final Clock clock;
+  const SearchRidePage({super.key, this.clock = const Clock()});
```
fynsta, line 21: Aaah I did not realise this would be this horrible...

MarkFalke:
When I searched for rides with the whole day option activated when I selected the next day the ride from the day before is still displayed. Is this behaviour right?

Eddie-42:
> When I searched for rides with the whole day option activated when I selected the next day the ride from the day before is still displayed. Is this behaviour right?

Depends. Right now we are checking if the start time OR end time of a drive fits, so if the start and end times are on different days, it's intended behavior. Was that the case?

MarkFalke:
It is a bit hard to explain, maybe I can just show you?

MarkFalke (approved):

PR #116, opened 2023-02-03
Fix script for fvm

fynsta:
Only really relevant for @istzen. This should add support for fvm in the coverage script.


Showcas (approved):

istzen (approved):
thanks

PR #117, opened 2023-02-05
Verify translations

fynsta:
Check that the file of untranslated messages is empty.


istzen:
aren't there some cases were we don't need translation? so the file is not always empty? or did we remove all those cases?

fynsta:
I'm not sure what you mean? Every string that is in the english arb file should also be in the german arb file (even if they are the same string, because they might not be the same string in other languages).

Eddie-42 (approved):

istzen (approved):

PR #118, opened 2023-02-06
Change increment field card to container

Eddie-42:
As discussed. I also added a little bit of padding on the SearchRidePage the card had a little bit of space around it, which was now missing.

### DIFF
--- a/lib/drives/pages/create_drive_page.dart
+++ b/lib/drives/pages/create_drive_page.dart
@@ -30,7 +30,7 @@ class _CreateDrivePageState extends State<CreateDrivePage> {
         title: Text(S.of(context).pageCreateDriveTitle),
       ),
       body: const Padding(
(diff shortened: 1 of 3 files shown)


fynsta:
I think we also wanted to remove at least one of the dividers?

fynsta (approved):

lib/drives/pages/create_drive_page.dart
```diff
@@ -240,7 +240,7 @@ class _CreateDriveFormState extends State<CreateDriveForm> {
               ),
             ],
           ),
-          const Divider(),
+          const SizedBox(height: 5),
```
istzen, line 243:
```suggestion
          const SizedBox(height: 10),
```
then it's the same as the others. 
Same for search

Showcas (approved):

MarkFalke (approved):

istzen (approved):

PR #122, opened 2023-02-06
Fix SVG warnings

fynsta:
Just cleaning up the SVG file a little bit to fix all the warnings that flutter always preduces when it shows that picture:
1) It does not know "mm" as a unit, so I just removed that (that should be pixels now, but it does not matter because we resize it anyways)
2) All the Inkscape autogenerated stuff cannot be parsed by flutter, so I just removed it.

### DIFF
--- a/assets/logo.svg
+++ b/assets/logo.svg
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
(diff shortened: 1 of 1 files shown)


fynsta:
Github does not show the diff for rich files, you can see the text diff by adding ".diff" to the URL like so:
https://github.com/MOTIS-Mitfahr-App/flutter-app/pull/122.diff

Eddie-42 (approved):

istzen (approved):

PR #128, opened 2023-02-09
Recurring drives

fynsta:
This is basically just the backend for later (SQL and Flutter models).

Please review especially the ideas in here. Especially the sql functions are pretty important. 

One thing that you cannot see in the dump: We added a cronjob that runs every night at 3 am and starts the create_drives_from_recurring function. That one creates a new drive in 30 days, if required by the recurrence rule.


supabase/pg_dump.sql
```diff
@@ -112,6 +1546,44 @@ end;$$;
 
 ALTER FUNCTION "public"."create_chat"() OWNER TO "postgres";
 
+--
+-- Name: create_drives_from_recurring(); Type: FUNCTION; Schema: public; Owner: postgres
+--
+
+CREATE FUNCTION "public"."create_drives_from_recurring"() RETURNS "void"
```
fynsta, line 1553: This one is important

supabase/pg_dump.sql
```diff
@@ -293,6 +1765,103 @@ end;$$;
 
 ALTER FUNCTION "public"."mark_ride_event_as_read"("ride_event_id" integer) OWNER TO "postgres";
 
+--
+-- Name: recurring_drive_insert(); Type: FUNCTION; Schema: public; Owner: postgres
+--
+
+CREATE FUNCTION "public"."recurring_drive_insert"() RETURNS "trigger"
```
fynsta, line 1772: this one is important too

supabase/pg_dump.sql
```diff
@@ -293,6 +1765,103 @@ end;$$;
...
+$$;
+
+
+ALTER FUNCTION "public"."recurring_drive_insert"() OWNER TO "postgres";
+
+--
+-- Name: recurring_drive_update(); Type: FUNCTION; Schema: public; Owner: postgres
+--
+
+CREATE FUNCTION "public"."recurring_drive_update"() RETURNS "trigger"
```
fynsta, line 1811: and this is the most important of all

fynsta:
I added comments for the most important parts of the SQL dump. Of course the dart files should be reviewed as normally.

lib/drives/models/recurring_drive.dart
```diff
@@ -0,0 +1,103 @@
...
+
+import '../../account/models/profile.dart';
+import '../../util/parse_helper.dart';
+import '../../util/search/position.dart';
+import '../../util/supabase_manager.dart';
+import '../../util/trip/trip_like.dart';
+import 'drive.dart';
+
+class RecurringDrive extends TripLike {
+  DateTime? stoppedAt;
```
Eddie-42, line 13: Why do we need the information when it was stopped and not only that is was stopped?

supabase/pg_dump.sql
```diff
@@ -112,6 +1546,44 @@ end;$$;
...
+	FOR recurring_drive IN
+		SELECT
+			*
+		FROM
+			recurring_drives
+		WHERE
+			_rrule.rruleset (recurrence_rule) @> (now()::timestamp + interval '30 days') AND
+			stopped_at IS NULL
+	LOOP
+		drive_start_time = CURRENT_DATE + interval '30 days' + recurring_drive.start_time;
```
Eddie-42, line 1570: Can you refactor this into a function?

supabase/pg_dump.sql
```diff
@@ -112,6 +1546,44 @@ end;$$;
...
+	drive_start_time TIMESTAMP;
+	drive_end_time TIMESTAMP;
+BEGIN
+	FOR recurring_drive IN
+		SELECT
+			*
+		FROM
+			recurring_drives
+		WHERE
+			_rrule.rruleset (recurrence_rule) @> (now()::timestamp + interval '30 days') AND
```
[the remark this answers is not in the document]
fynsta: Done, I don't like the syntax too much but fine. (Although you need to consider that we need to write custom logic when changing this value anyways -> for example we need to fill the gap with trips if we make that interval longer)

supabase/pg_dump.sql
fynsta: I certainly can, the question is whether it is a good idea. The function would look like this

```
CREATE OR REPLACE FUNCTION public.recurring_drive_generate_datetimes(base_date date, start_time time without time zone, end_time time without time zone)
 RETURNS record
 LANGUAGE plpgsql
 IMMUTABLE
AS $function$
DECLARE
	ret RECORD;
	generated_start_time TIMESTAMP;
	generated_end_time TIMESTAMP;
BEGIN
   	generated_start_time = base_date + start_time;
	generated_end_time = base_date + end_time;
	IF start_time > end_time THEN
		generated_end_time = generated_end_time + interval '1 day';
	END IF;
	
	SELECT generated_start_time, generated_end_time INTO ret;
	
	RETURN ret;	
END; $function$
```

and then it would be called like this
```
SELECT generated_start_time, generated_end_time 
INTO drive_start_time, drive_end_time 
FROM recurring_drive_generate_datetimes(base_date, recurring_drive.start_time, recurring_drive.end_time) 
AS ret (generated_start_time TIMESTAMP, generated_end_time TIMESTAMP);
```

That is compared to this code that was used before
```
drive_start_time = next_date + recurring_drive.start_time;
drive_end_time = next_date + recurring_drive.end_time;
IF recurring_drive.start_time > recurring_drive.end_time THEN
	drive_end_time = drive_end_time + interval '1 day';
END IF;
```

lib/drives/models/recurring_drive.dart
fynsta: I thought we might add the option to say "Stop next week"/"Stop at specific date" or something like that? Otherwise there is no way to stop the recurring drive without having a drive scheduled literally the next day, where people might already ride with you.

And if not, this is still good practice because it stores interesting information for us.

supabase/pg_dump.sql
fynsta: Well now that I have already written it, I will (tentatively) add it

fynsta:
This is part 1 of recurring drives, but Flutter stuff and layout will come in a second PR to make sure this is not too big. So please review this _as is_, and keep in mind that this is really just the backend the the upcoming frontend logic will depend on.

Eddie-42 (approved):

lib/drives/models/recurring_drive.dart
Eddie-42: Okay but you didn't actually implement that in the Postgres functions correctly I think

supabase/pg_dump.sql
Eddie-42: What I had in mind was to include the INSERT in the function as well

lib/drives/models/recurring_drive.dart
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @override
+  Map<String, dynamic> toJson() {
+    return super.toJson()
+      ..addAll(<String, dynamic>{
+        'stopped_at': stoppedAt?.toString(),
+        'start_time': startTime.formatted,
+        'end_time': endTime.formatted,
+        'recurrence_rule': 'DTSTART:${(createdAt ?? DateTime.now()).millisecondsSinceEpoch}\n$recurrenceRule',
```
Eddie-42, line 72: Here we discussed to allow shenanigans like pausing a recurrence rule by setting the start time to something else right? I still don't know if within the recurrence rule is the right place to do that or if it should just be another column

fynsta: I'll have to think about that some more. But I don't think that should be part of the first PR for this feature, it's more of an auxiliary feature building on that one.

lib/drives/models/recurring_drive.dart
fynsta: I think I did that correctly live? But yeah you are right, and I will double-check

istzen (approved):

MarkFalke (approved):

PR #134, opened 2023-02-16
Card notification icons

Showcas:
adds notification icons for the cards in the top right

the price is only shown on the preview cards for the search

are you happy with the icon choices?

### DIFF
--- a/lib/util/trip/drive_card.dart
+++ b/lib/util/trip/drive_card.dart
@@ -55,7 +55,6 @@ class DriveCardState extends TripCardState<Drive, DriveCard> {
     if (mounted) {
       setState(() {
         _fullyLoaded = true;
-        statusColor = pickStatusColor();
       });
     }
   }
@@ -83,20 +82,20 @@ class DriveCardState extends TripCardState<Drive, DriveCard> {
 
   @override
   Widget buildTopRight() {
-    if (!_fullyLoaded) {
+    if (!_fullyLoaded || _drive.isFinished) {
       return const Center(
         child: SizedBox(),
       );
     } else if (_drive.cancelled) {
       return Icon(
         Icons.block,
-        color: statusColor,
+        color: getStatusColor,
         key: const Key('cancelledIcon'),
       );
     } else if (_drive.rides!.any((Ride ride) => ride.status == RideStatus.pending)) {
       return Icon(
         Icons.access_time_outlined,
-        color: statusColor,
+        color: getStatusColor,
         key: const Key('pendingIcon'),
       );
     } else {
@@ -107,12 +106,10 @@ class DriveCardState extends TripCardState<Drive, DriveCard> {
   }
 
   @override
-  Color pickStatusColor() {
+  Color get getStatusColor {
     if (!_fullyLoaded) {
       return Theme.of(context).cardColor;
-    } else if (drive.status == DriveStatus.preview) {
-      return Theme.of(context).colorScheme.primary;
-    } else if (drive.endDateTime.isBefore(DateTime.now())) {
+    } else if (_drive.isFinished) {
       return Theme.of(context).disabledColor;
     } else {
       if (drive.status.isCancelled()) {
--- a/lib/util/trip/drive_card.dart
+++ b/lib/util/trip/drive_card.dart
@@ -91,16 +91,19 @@ class DriveCardState extends TripCardState<Drive, DriveCard> {
       return Icon(
         Icons.block,
         color: pickStatusColor(),
+        key: const Key('cancelled'),
       );
     } else if (_drive.rides!.any((Ride ride) => ride.status == RideStatus.pending)) {
       return Icon(
         Icons.access_time_outlined,
         color: pickStatusColor(),
+        key: const Key('pending'),
       );
     } else {
       return Icon(
         Icons.done_all,
         color: pickStatusColor(),
+        key: const Key('no pending'),
       );
     }
   }
--- a/lib/util/trip/drive_card.dart
+++ b/lib/util/trip/drive_card.dart
@@ -23,7 +23,6 @@ class DriveCardState extends TripCardState<Drive, DriveCard> {
   @override
   void initState() {
     super.initState();
-
     setState(() {
       drive = widget.trip;
       trip = drive;
@@ -56,6 +55,7 @@ class DriveCardState extends TripCardState<Drive, DriveCard> {
     if (mounted) {
       setState(() {
         _fullyLoaded = true;
+        statusColor = pickStatusColor();
       });
     }
   }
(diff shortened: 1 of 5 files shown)


fynsta:
I don't love the way it looks (and the way the information is just redundant), but I cannot come up with a better way right now... The icons are fine, but they just feel a little out of place.

lib/util/trip/ride_card.dart
```diff
@@ -92,7 +92,42 @@ class _RideCardState extends TripCardState<Ride, RideCard> {
...
+          );
+        case RideStatus.approved:
+          return Icon(
+            Icons.done_all,
+            color: pickStatusColor(),
+            key: const Key('approved'),
+          );
+        case RideStatus.rejected:
+          return Icon(
+            Icons.close,
```
fynsta, line 117: Icons.block makes more sense here, and use something else for cancellation?

lib/util/trip/drive_card.dart
```diff
@@ -72,6 +72,33 @@ class _DriveCardState extends TripCardState<Drive, DriveCard> {
...
+  Widget buildTopRight() {
+    if (!_fullyLoaded) {
+      return const Center(
+        child: SizedBox(),
+      );
+    } else if (_drive.cancelled) {
+      return Icon(
+        Icons.block,
+        color: pickStatusColor(),
+        key: const Key('cancelled'),
```
fynsta, line 85: the key should contain the fact that this is a statusindicator

test/widgets/cards_test.dart
```diff
@@ -259,9 +259,6 @@ void main() {
 
       //Rating is shown
       expect(find.byType(CustomRatingBarIndicator), findsOneWidget);
-
-      //Price is shown
-      expect(find.byKey(const Key('price')), findsOneWidget);
```
fynsta, line 264: this seems unrelated?

test/widgets/cards_test.dart
```diff
@@ -420,26 +431,30 @@ void main() {
...
         await loadDriveCard(tester, drive);
 
+        expect(find.byKey(const Key('cancelled')), findsOneWidget);
+
         final BuildContext context = tester.element(find.byType(Container).first);
         expect(tester.widget<Card>(find.byType(Card)).color, Theme.of(context).colorScheme.error);
       });
 
-      testWidgets('waring color for drive with ride requests', (WidgetTester tester) async {
+      testWidgets('waring status for drive with ride requests', (WidgetTester tester) async {
```
fynsta, line 448: ofc not related to your PR, but I just noticed: waring -> warning

test/widgets/cards_test.dart
Showcas: the price was before there where the icon was and the price is only shown for preview status

fynsta: ah yes, thanks!

lib/util/trip/ride_card.dart
Showcas: I think its easiest for all cancelled/rejected rides to have the same icon

lib/util/trip/drive_card.dart
```diff
@@ -72,6 +72,33 @@ class _DriveCardState extends TripCardState<Drive, DriveCard> {
     return SeatIndicator(trip);
   }
 
+  @override
+  Widget buildTopRight() {
+    if (!_fullyLoaded) {
+      return const Center(
```
fynsta, line 78: what does the Center do?

lib/util/trip/drive_card.dart
```diff
@@ -72,6 +72,33 @@ class _DriveCardState extends TripCardState<Drive, DriveCard> {
...
+        child: SizedBox(),
+      );
+    } else if (_drive.cancelled) {
+      return Icon(
+        Icons.block,
+        color: pickStatusColor(),
```
fynsta, line 84: let's load the color into a variable, we need it in all three cases

lib/util/trip/drive_card.dart
Showcas: it can be taken out, mb

MarkFalke (approved):

istzen (approved):
i aggre with @fynsta, but apart from that everything is fine

lib/util/trip/drive_card.dart
```diff
@@ -72,6 +72,33 @@ class _DriveCardState extends TripCardState<Drive, DriveCard> {
...
+        key: const Key('cancelledIcon'),
+      );
+    } else if (_drive.rides!.any((Ride ride) => ride.status == RideStatus.pending)) {
+      return Icon(
+        Icons.access_time_outlined,
+        color: statusColor,
+        key: const Key('pendingIcon'),
+      );
+    } else {
+      return Icon(
```
Eddie-42, line 94: IMO, you don't actually need an icon in this case. We don't want to give the impression that the drive is somehow "finished".

Eddie-42 (commented):
Sorry for the late review. Otherwise looking good!

lib/util/trip/ride_card.dart
```diff
@@ -92,7 +93,37 @@ class _RideCardState extends TripCardState<Ride, RideCard> {
 
   @override
   Widget buildTopRight() {
-    return Text(key: const Key('price'), ' ${_ride.price}€');
+    if (!_fullyLoaded) {
```
Eddie-42, line 96: This also applies here, and why did we decide to make past rides green depending on whether they are approved? I think they should be grey. If I remember correctly, you can then still tell if a ride is rejected/cancelled because the whole card is grey.

Showcas: past drives that took place are green, past drives taht did not are grey (cancelled)

Showcas: To me it is just to make it more understandable, but yea... you think all grey and greyout the ones that did not take place?

lib/util/trip/drive_card.dart
```diff
@@ -72,6 +72,31 @@ class _DriveCardState extends TripCardState<Drive, DriveCard> {
...
+        child: SizedBox(),
+      );
+    } else if (_drive.cancelled) {
+      return Icon(
+        Icons.block,
+        color: statusColor,
+        key: const Key('cancelledIcon'),
+      );
+    } else if (_drive.rides!.any((Ride ride) => ride.status == RideStatus.pending)) {
+      return Icon(
```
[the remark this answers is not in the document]
Showcas: But shouldn't the notification icons be an additional idicator. like yes I understnd that it is "trivial". But at least put an icon for drive/rides that took place and for the ones that didn't

lib/util/trip/ride_card.dart
fynsta: Yes I think so too. Makes it more consistent with Drive

lib/util/trip/drive_card.dart
```diff
@@ -72,6 +72,31 @@ class _DriveCardState extends TripCardState<Drive, DriveCard> {
...
+      );
+    } else {
+      return const Center(
+        child: SizedBox(),
+      );
+    }
+  }
+
   @override
   Color pickStatusColor() {
```
fynsta, line 101: We talked about this but please make this a getter function

lib/util/trip/drive_card.dart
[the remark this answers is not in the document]
fynsta: What do you care if there were pending rides for past drives? They are irrelevant

lib/util/trip/ride_card.dart
fynsta: I would say they are basically rejected from a visual standpoint

Eddie-42 (approved):

PR #135, opened 2023-02-22
Recurring drives frontend

fynsta:
We're truly sorry, this is the biggest feature by far. This is the front end part of the recurring drives!
Instead of all the new changes, I'm going to describe a flow for testing:
1. You can create recurring drives by clicking the checkbox on the regular create drive page. Please check if everything is understandable for usability.
2. You are taken to the new recurring drive detail page, where you can see your schedule, information and all generated drives. In addition to that, you can edit the schedule by tapping the icon on the top right.
3. You can see your recurring drives on the new "Recurring" tab under the Drives navigator bar entry.
The rest of the time was essentially spent on figuring out edge cases and weirdness that has to do with the complexity of situations that can arise by cancelling, changing or picking weird recurring drive schedules.

We haven't finished everything (especially tests), but please review what we have until now!


Eddie-42 (approved):

MarkFalke:
The option where you can see the next few upcomming drives as a preview is only available when the next drives are during the next 30 days. Is this right or should we show the next 2 dates even if they are a year apart?

fynsta:
> The option where you can see the next few upcomming drives as a preview is only available when the next drives are during the next 30 days. Is this right or should we show the next 2 dates even if they are a year apart?

Yes that might be a little confusing... We can't show any other drives because they were not created yet by the cronjob. Theoretically I could of course lengthen that timeframe. But that would require cancelling many more drives when changing the recurrence rule.

I think 30 days is a good medium timelength, but we should think of some way to better visualize/document this for the end user.

MarkFalke (approved):

istzen:
when you change the language to german the "wiederkehrende" is not completely visible

istzen:
the following exception occurs when i try to stop a drive. 
`[VERBOSE-2:dart_vm_initializer.cc(41)] Unhandled Exception: PostgrestException(message: column "stopped_at" does not exist, code: 42703, details: Bad Request, hint: null)
#0      PostgrestBuilder._parseResponse (package:postgrest/src/postgrest_builder.dart:323:7)
#1      PostgrestBuilder._execute (package:postgrest/src/postgrest_builder.dart:198:14)
<asynchronous suspension>
#2      PostgrestBuilder.then (package:postgrest/src/postgrest_builder.dart:400:24)
<asynchronous suspension>`

test/widgets/recurring_drive_detail_page_test.dart
```diff
@@ -0,0 +1,160 @@
...
+        await pumpMaterial(tester, RecurringDriveDetailPage.fromRecurringDrive(recurringDrive));
+        await tester.pump();
+
+        final driveCardFinder = find.byType(DriveCard, skipOffstage: false);
+        expect(driveCardFinder, findsNWidgets(RecurringDriveDetailPageState.shownDrivesCountDefault));
+        final driveCardState = tester.state<DriveCardState>(driveCardFinder.first);
+        expect(driveCardState.trip.id, equals(recurringDrive.drives!.first.id));
+      });
+
+      testWidgets('Can show less and more upcoming drives', (WidgetTester tester) async {
```
istzen, line 102: this test failed, for me.

fynsta:
> when you change the language to german the "wiederkehrende" is not completely visible

You mean in the tabview? Yeah okay but what can you do 🤷‍♂️

test/widgets/recurring_drive_detail_page_test.dart
fynsta: It also seems to have failed in the build pipeline on the last commit, so that test is probably flaky. I'll look into it

fynsta:
> the following exception occurs when i try to stop a drive. `[VERBOSE-2:dart_vm_initializer.cc(41)] Unhandled Exception: PostgrestException(message: column "stopped_at" does not exist, code: 42703, details: Bad Request, hint: null) #0 PostgrestBuilder._parseResponse (package:postgrest/src/postgrest_builder.dart:323:7) #1 PostgrestBuilder._execute (package:postgrest/src/postgrest_builder.dart:198:14) <asynchronous suspension> #2 PostgrestBuilder.then (package:postgrest/src/postgrest_builder.dart:400:24) <asynchronous suspension>`

Whoops. Should be fixed now?

Showcas (approved):

istzen:
i have an overflow error, at the elements that show on which days the drive will be.

istzen:
and i can not stop the recurring drive.

PR #136, opened 2023-02-23
Flexible

fynsta:
Make some stuff more flexible using overflows and wraparounds. Currently I fixed
- TripCard start/end
- ProfileWidget
- ProfileChip

If you have any other overflow errors or something with my changes does not work, please let me know. You should test this by using longer strings in the db for the concerned fields (for example I have chosen a name with 30 characters to check that everything works).


lib/drives/pages/drive_detail_page.dart
```diff
@@ -354,29 +356,21 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
-                            style: const TextStyle(color: Colors.white),
-                            textScaleFactor: 1.0,
-                          ),
-                          showBadge: action.ride.chat!.getUnreadMessagesCount() != 0,
-                          position: badges.BadgePosition.topEnd(top: -12),
-                          child: const Icon(Icons.chat),
-                        ),
+                    IconWidget(icon: icon, count: action.ride.seats),
+                    const SizedBox(width: 20),
+                    // TODO: Center this again
```
fynsta, line 362: @Eddie-42 I might need your help here. I remember we spent a long time trying to get this to the center. Now I had to rip that open again because the ProfileWidget itself needs to be Flexible.

Now it looks like this: (left-aligned but with the padding of 20 pixels).
image
But I could not figure out how to center it inside of the Flexible/Expanded. Do we want to leave it like this because that page in particular will be redesigned anyways?

lib/account/widgets/review_detail.dart
```diff
@@ -15,20 +15,22 @@ class ReviewDetail extends StatelessWidget {
...
-          style: TextStyle(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5)),
-        ),
         Expanded(
-          child: Align(
-            alignment: Alignment.centerRight,
-            child: CustomRatingBarIndicator(
-              rating: review.rating.toDouble(),
-              size: CustomRatingBarSize.medium,
-            ),
+          child: Row(
```
Showcas, line 19: is a Row in a row not the same as having the elements of the inner row as normal ones of the outer?

Showcas (approved):

MarkFalke (approved):

istzen (approved):

PR #137, opened 2023-02-26
Fix upload button permission handling

fynsta:
Does a few things:

- Fixes bug where circle icon would vanish on click (the function should not be null, we can just return at the beginning of the function instead)
- Adds some shadow to the circle icon to make it stand out better. How do you think it looks?
- Proposes to go to app settings when the photos permission is not given (weirdly, the ImagePicker does not use any permission, I have no idea how that even works?).

Please try it out on both iOS and Android (those who can do both, anyways). You may need to reset your permissions or create a new emulator device to recreate the situations where it will actually ask you for permission, because you have probably given permission previously.


istzen (approved):
looks good

MarkFalke (approved):

Showcas (approved):

PR #138, opened 2023-02-26
Improve reviews

fynsta:
Two things:
1. Fixes bug where review date would not update when you changed the review (it always showed createdAt, but that makes no sense when you change your review).
2. Finally adds state to the ReviewDetail, so that the first 2 lines (do we want to show more?) are shown, and then you click on "Show More" to reveal the rest.


istzen:
the review gets cut off but i can't see the whole review (the show more and show less buttons are either not shown or not working correctly)

fynsta:
Can you send a picture? I never had that

fynsta:
Okay I removed the "..." overflow indicator because that caused the weird behaviour that you observed

test/widgets/review/review_detail_test.dart
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    testWidgets('Works without text', (WidgetTester tester) async {
+      await pumpReviewDetail(tester, ReviewFactory().generateFake(text: NullableParameter(null)));
+      expect(find.byType(ProfileChip), findsOneWidget);
+      expect(find.byType(CustomRatingBarIndicator), findsOneWidget);
+      expect(expandReviewButtonFinder, findsNothing);
+    });
+
+    testWidgets('Can expand and retract when text is long', skip: true, (WidgetTester tester) async {
```
fynsta, line 48: @Eddie-42 maybe you can look into this? This test does not work because the LayoutBuilder does not realize that the text would exceed two lines. No idea why

Eddie-42 (approved):

MarkFalke (approved):

istzen (approved):
looks all good now. What could be added is that you can see the category ratings for each user. But i don't think we have the time to do that

fynsta:
> looks all good now. What could be added is that you can see the category ratings for each user. But i don't think we have the time to do that

Yes I thought about that too but I think the UI would get really cluttered and confusing with all the Rating Bars

PR #142, opened 2023-03-04
Redesign home page

Eddie-42:
As promised, I redesigned the home page!
* I divided the "notifications" area into 3 parts. To do that, I split the all-encompassing `_items` list into what I consider its natural 3 parts (upcoming trips, ride events, and messages). I also reordered them in the code to reflect what will be shown on screen, so that's probably the majority of what you see in the diff :D
* I added a new notification: Complete profile! It tells the user that they're "very mysterious" and there is a button to the profile page. It appears if no other notifications are present (If the user has "nothing else to do"), and the user hasn't configured any personal info except features (I didn't consider features personal information, but we can debate that). I am aware that the page doesn't update if you change your profile via the normal way. I think if we want to implement that, we'll have to add a listener and notify them every time currentProfile changes. I didn't consider that vital, but if you want me to do that, let me know :)


lib/home_page.dart
```diff
@@ -175,172 +155,220 @@ class HomePageState extends State<HomePage> {
...
+          ),
+        );
+      }
+      if (notifications.isEmpty) {
+        if (supabaseManager.currentProfile!.description == null &&
+            supabaseManager.currentProfile!.birthDate == null &&
+            supabaseManager.currentProfile!.surname == null &&
+            supabaseManager.currentProfile!.name == null &&
+            supabaseManager.currentProfile!.gender == null &&
+            supabaseManager.currentProfile!.avatarUrl == null) {
```
fynsta, line 265: let's move that into a method

lib/home_page.dart
```diff
@@ -175,172 +155,220 @@ class HomePageState extends State<HomePage> {
...
+      if (_trips.isNotEmpty) {
+        notifications.add(
+          Column(
+            key: const Key('tripsColumn'),
+            children: <Widget>[
+              Align(
+                alignment: Alignment.centerLeft,
+                child: Text(S.of(context).pageHomePageTrips, style: Theme.of(context).textTheme.titleLarge),
+              ),
+              const SizedBox(height: 10),
```
fynsta, line 223: can we have a little more vertical space... the categories are sticking together a little too much for my taste

fynsta:
@Eddie-42 I don't know if it's possible but there is one thing I think would be really cool: The buttons on the HomePage should all switch you to the correct navigator and then open the respective page there. Because the HomePage in itself should always be the one page shown on the Home tab. And all the buttons have different semantics - some are connected to drives, some to rides, some to your account. It would make perfect sense to me

Eddie-42:
I think that's possible, but I haven't gotten it to work for like 10 min and I need my sleep. Will try again later!

lib/account/models/profile.dart
```diff
@@ -51,6 +51,14 @@ class Profile extends Model {
 
   bool get isCurrentUser => id == supabaseManager.currentProfile?.id;
 
+  bool get hasNoPersonalInformation =>
```
fynsta, line 54: tests for this? :D A case for Copilotx

Eddie-42:
> The buttons on the HomePage should all switch you to the correct navigator and then open the respective page there.

Done! The only problem is that you now need a MainApp for the tests, but I'll figure that out another time

Showcas (approved):

PR #143, opened 2023-03-04
Redesign trip overview

Eddie-42:
You can consider this Part I of the Detail Pages redesign. It's kind of more of a personal project because I always considered this the very ugliest part of the app (I particularly disliked how small and disconnected start and destination were: You'd probably never notice it at all since there are bigger, more colorful elements underneath - and if you did, you'd not get the connection right away).

I tried to fix that for the Drive Overview by making start and end bigger and bolder while making the times less bold and redesigning the rest of the info to be around, not strictly underneath that.

And yes, you heard correctly: There is now a Drive Overview, which means there is also a Ride Overview. The reason for the start-destination part of the overview not being a timeline is that we felt it looked weird on the Drive Detail Page (where there is a basically identical timeline underneath it, containing the stops and where to pick people up and drop them off). However, I split the Overviews because I still think it makes more sense on the Ride Detail Page to include a proper timeline, like you would see on the cards. If you don't like it, I can revert that. Also, I included the price in the Ride Overview.

Speaking of cards, I redesigned them just a smidge to make them consistent with the boldness and sizes of the start and destination texts of the Drive Overview. A side effect is that they now contain just a little less dead space, which was intentional.


fynsta:
Wow that text is insanely huge now. 
image
When it already wraps after one word on the big emulator device I'm using - This might break complely on smaller devices.

fynsta:
Also, why is it "Date on the left - stuff - seats on the left"? Very asymmetrical and random

fynsta:
We already talked about this, but I don't like the visual inconsistency of having different designs for overviews... I guess there's no way around it if you hate the old one this strongly for some reason.

I assume recurring drives will use the DriveOverview? But have fun rebasing that.

Eddie-42:
The text is quite big on purpose. As I already talked about, this is to 1) make it stand out as the most important part of the page (which it is: Your eyes should first go to the overview and then to the detailed route plan) despite it not being as colorful or in the middle and 2) make the connection between the start and destination clearer: Before, the whole page looked like a row of columns because there were two elements that were near to each other on the left and on the right side and a big empty space in between. The bigger text size helps make the space smaller. (That is also why I put the date and seats on the left, because now it looks like a column with a row in it. I also tried to center date and seats and that looks bad.)

However, I admit that there isn't much room for long place names on your screen. Are you testing that on an iPhone? Because on my Pixel 6 emulator, I think things fit quite nicely. In any way, I don't think it's a huge problem because 1) As we have already noticed multiple times I think, the iPhone emulator screen which I suspect you're using is very small compared to even most screens out in the wild, 2) there aren't that many very long names and we can't make it look nice for every length, so I figure we should optimize the design for rather short ones (especially because the very long ones won't break the app, just look slightly worse), and 3) There is a huge gap between Flutter's built-in titleLarge, which I used, and titleMedium, which feels like half the size and isn't really bold and I would like to use standard designs because I deem them to have the highest longevity.

Eddie-42:
For the two overviews, maybe a little survey in the next meeting? I'm not hugely passionate about it, but I do think the timelines variant looks nicer for the Ride Overview.

Yes, for the Recurring Drives I think we could just use the DriveOverview with little adjustments.

fynsta:
> However, I admit that there isn't much room for long place names on your screen. Are you testing that on an iPhone? Because on my Pixel 6 emulator, I think things fit quite nicely.

This is how it looks on a Pixel 5 which is still a relatively big phone compared to others that exist in the wild.
image
The text should not have to wrap into three lines for such a normal name. It just does not look good. I'm so paranoid because of the image that Felix sent me. I don't know exactly which device he has, but it is still a lot smaller than the Pixel 5 (I know that because I can see where it cut off the "Wir empfehlen deinen echten Namen" way earlier). We _should not_ only design for big screen sizes, that would be an easy mistake to make. 

I think the huge text is fine on the cards because there is more horizontal space. But in the DriveOverview there just isn't that kind of space.

I understand the huge gap, but I still think we should use the smaller font.

Eddie-42:
We can discuss that in a meeting.

lib/util/trip/pending_ride_card.dart
```diff
@@ -66,13 +67,13 @@ class _PendingRideCardState extends TripCardState<Ride, PendingRideCard> {
...
           onPressed: () => showApproveDialog(context),
-          icon: const Icon(Icons.check_circle_outline, color: Colors.green, size: 50.0),
+          icon: Icon(Icons.check_circle_outline, color: Theme.of(context).own().success, size: 42.0),
           tooltip: S.of(context).approve,
         ),
         IconButton(
           key: const Key('rejectButton'),
           onPressed: () => showRejectDialog(context),
-          icon: const Icon(Icons.cancel_outlined, color: Colors.red, size: 50.0),
+          icon: Icon(Icons.cancel_outlined, color: Theme.of(context).colorScheme.error, size: 42.0),
```
istzen, line 76: i liked the icons better with the bigger size, what do the others think?

istzen:
i aggree with @fynsta the textsize is maybe a little bit to big, also the words a somtimes cut off at weird places. i had something where just the last letter was in a new line. I understand the idea that the overview should be the first thing you look at, but that's more important for the rides than for the drives. I think since there you have created it yourself and also have the timeline below with more information, it's okay if there the text is smaller. 
For the RideOverview i agree with @Eddie-42, i like the timeline in think it's better that way because before the focus was to much on the Reviews. And i don't think the different design is a problem, after all rides and drives are different things

Showcas (approved):

fynsta (approved):

PR #144, opened 2023-03-04
Redesign drive detail page

Eddie-42:
Finally, a redesign of the drives detail page! I've never been a fan of the containers with a border, and thought it much more practical to use what I think Flutter intends for this use case: A card. I think this worked out quite well. Otherwise, I made a few changes to font sizes, spacing etc here and there.

It is worth mentioning that I included a few headers like "Driver" on the Ride Detail Page or "Route" on the Drive Detail Page. I think this looks better than Dividers. It is also a response to the user studies, where some users didn't get that the features belonged to the drivers, or why the start and destination information is displayed twice on the Drive Detail Page.

### DIFF
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -155,7 +155,7 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
             return Padding(
               padding: const EdgeInsets.only(left: 8.0),
               child: Column(
-                children: <Widget>[const SizedBox(height: 5), buildCard(stop), const SizedBox(height: 5)],
+                children: <Widget>[Padding(padding: const EdgeInsets.symmetric(vertical: 5), child: buildCard(stop))],
               ),
             );
           },
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -179,12 +179,13 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
           itemCount: stops.length,
         ),
       );
-      widgets.add(timeline);
-
       if (approvedRides.isNotEmpty) {
+        widgets.add(timeline);
+
         widgets.add(const Divider(thickness: 1));
         final Set<Profile> riders = approvedRides.map((Ride ride) => ride.rider!).toSet();
         widgets.add(ProfileWrapList(riders, title: S.of(context).riders));
+        widgets.add(const Divider(thickness: 1));
       }
 
       final List<Ride> pendingRides = _drive!.pendingRides!.toList();
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -174,7 +174,6 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
           timeline,
           const Divider(),
           ProfileWrapList(riders, title: S.of(context).riders),
-          const Divider(),
         ]);
       }
 
--- a/lib/drives/pages/drive_detail_page.dart
+++ b/lib/drives/pages/drive_detail_page.dart
@@ -155,24 +155,7 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
             return Padding(
               padding: const EdgeInsets.only(left: 8.0),
               child: Column(
-                children: <Widget>[
-                  const SizedBox(height: 10.0),
-                  Container(
-                    decoration: BoxDecoration(
-                      border: Border.all(color: Theme.of(context).colorScheme.onSurface.withOpacity(0.5)),
-                      borderRadius: BorderRadius.circular(5.0),
-                    ),
-                    padding: const EdgeInsets.all(8.0),
-                    width: double.infinity,
-                    child: Column(
-                      key: const Key('waypointCard'),
-                      mainAxisSize: MainAxisSize.min,
-                      crossAxisAlignment: CrossAxisAlignment.start,
-                      children: buildCard(stop),
-                    ),
-                  ),
-                  const SizedBox(height: 10.0)
-                ],
+                children: <Widget>[const SizedBox(height: 5), buildCard(stop), const SizedBox(height: 5)],
               ),
             );
           },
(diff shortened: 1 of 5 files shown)


fynsta:
Tests are failing

lib/rides/pages/ride_detail_page.dart
```diff
@@ -113,13 +113,30 @@ class _RideDetailPageState extends State<RideDetailPage> {
...
-
-      widgets.add(ReviewsPreview(driver));
-
-      if (driver.profileFeatures!.isNotEmpty) widgets.add(const Divider(thickness: 1));
-      widgets.add(FeaturesColumn(driver.profileFeatures!));
+      widgets.addAll(<Widget>[
+        Align(
+          alignment: Alignment.centerLeft,
+          child: Text(S.of(context).pageRideDetailDriver, style: Theme.of(context).textTheme.titleLarge),
+        ),
```
fynsta, line 120: I think this one is unnecessary, the driver is already on the ride card and there we don't need to explain it either. But maybe put the divider below this row instead?

lib/rides/pages/ride_detail_page.dart
```diff
@@ -113,13 +113,30 @@ class _RideDetailPageState extends State<RideDetailPage> {
...
+            alignment: Alignment.centerLeft,
+            child: Text(S.of(context).pageProfileFeaturesTitle, style: Theme.of(context).textTheme.titleLarge),
+          ),
+          const SizedBox(height: 10),
+          FeaturesColumn(driver.profileFeatures!),
+        ],
+      ]);
 
       if (ride.status.isRealRider()) {
         widgets.add(const Divider(thickness: 1));
```
fynsta, line 142: Let's remove this divider then

fynsta:
Wasn't the whole point of the AG to add more space? You did not really do that, did you? For example more vertical padding between the headers

lib/rides/pages/ride_detail_page.dart
fynsta: Actually let's just add space everywhere instead of dividers?

lib/drives/pages/drive_detail_page.dart
```diff
@@ -155,36 +155,27 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
-                    child: Column(
-                      key: const Key('waypointCard'),
-                      mainAxisSize: MainAxisSize.min,
-                      crossAxisAlignment: CrossAxisAlignment.start,
-                      children: buildCard(stop),
-                    ),
-                  ),
-                  const SizedBox(height: 10.0)
-                ],
+                children: <Widget>[const SizedBox(height: 5), buildCard(stop), const SizedBox(height: 5)],
```
fynsta, line 158: Padding?

fynsta:
The problem with the Card widget in Flutter for me is that is just really hard to differentiate from the background. Which makes the whole page difficult to parse for me. Maybe we can change the default card color a little bit?

lib/rides/pages/ride_detail_page.dart
Eddie-42: 1) This is a detail page, so there may be a bit more detail for people who don't understand the profile on the card. 
2) I think the profile widget really doesn't have the context to make it work like on the cards, where there is only the most vital information displayed. Here, e.g. the riders are also displayed and there is much more information surrounding the widget, so I think it's less clear: If every rider gets a widget, maybe this belongs to the reviews right below? You might very well still guess correctly, but there are more possibilities here. 
3) It adds consistency and the widget felt strange being the only one without a title (except for the overview). Especially since the section "riders" has a title, "driver" should too.
4) It doubles as a title to the whole section marked by the dividers, where everything is details about the driver, including the reviews and features. This is in response to the user studies, where some people thought e.g. the features were your features. This is also why I don't think removing the divider below this is a good idea.

Eddie-42:
> Wasn't the whole point of the AG to add more space? You did not really do that, did you? For example more vertical padding between the headers

I think what AG really wanted was to make the app feel less cluttered. More padding isn't necessarily good if it is inconstistent thoughout the app - then, it can actually make it feel more cluttered. That's part of why I chose the space around the titles: It's consistent with the profile page, which I thought we agreed looked quite well-spaced.

lib/drives/pages/drive_detail_page.dart
Eddie-42: This is actually a good example where I think I accomplished the goal of making the app feel less cluttered without increasing the padding. With the cards instead of containers with borders, it feels (at least to me) like there is just one less element on the page, and I think this looks quite balanced overall now.

Eddie-42:
I'm going to trust you on colors because I think the cards look quite nice :D Also, I think it would be weird if the default looked this bad so maybe it looks better on an actual phone?

lib/drives/pages/drive_detail_page.dart
fynsta: No I mean why did you use SizedBoxes in a Column instead of Padding?

fynsta:
> More padding isn't necessarily good if it is inconstistent thoughout the app

Well yes, that's what I thought you would do when redesigning. Do a pass over the whole app and add consistent spacing. Not add a new design now that we and the AG are used to and happy with the original design.

lib/rides/pages/ride_detail_page.dart
fynsta: I thought we agreed that a well-designed app does not really need dividers, because things are presented and spaced in such a way that it guides your sight and understanding of how everything fits together.

Eddie-42: But dividers help guide your sight :D BlaBlaCar also has dividers and I thought we agreed that was a very well-designed app. I think dividers can be very useful to group things together, and as long as they are not overused, I don't think they help clarity

Eddie-42:
Obviously AG wasn't happy with the old design and I wasn't either. As I learned from GMTK, if you receive feedback, it is the job of a designer first to question what the actual problem is and that it is rarely what the users tell you it should be. In our case, when AG said we need to add more spacing, I think the actual problem was that the pages were too cluttered. Also, if you want to redesign the drive detail page using just spacing, you would have to add spacing outside AND inside the borders of the container, and then you would very quickly run into problems with small screen sizes.

istzen:
i have to agree with @fynsta, regarding the colors of the cards. Especially in the light mode it's hard to differentiate between card an background, so we should maybe change the color here.
I like the headers, but for the rideDetailPage maybe that should stand out more.

Showcas (approved):

MarkFalke (approved):

fynsta (approved):

PR #148, opened 2023-03-12
Block user

istzen:
This is the flutter PR for the blocking of a User. The only thing that happens here is that a user that is blocked can't sign in anymore. The Important part of this functionality will be in the corresponding admin PR, which can be found [here](https://github.com/MOTIS-Mitfahr-App/react-web/pull/6).

### DIFF
--- a/lib/l10n/app_de.arb
+++ b/lib/l10n/app_de.arb
@@ -24,6 +24,8 @@
   "delete": "Löschen",
   "preview": "Vorschau",
   "failureSnackBar": "Es ist ein Fehler aufgetreten.",
+  "blockedSnackBar": "Dein Account wurde geblockt, bitte kontaktiere uns für mehr information",
+
 
   "riders": "Mitfahrer*innen",
   "seats": "Plätze",
--- a/lib/l10n/app_de.arb
+++ b/lib/l10n/app_de.arb
@@ -24,7 +24,7 @@
   "delete": "Löschen",
   "preview": "Vorschau",
   "failureSnackBar": "Es ist ein Fehler aufgetreten.",
-  "blockedSnackBar": "Dein Account wurde geblockt, bitte kontaktiere uns für mehr information",
+  "blockedSnackBar": "Dein Account wurde geblockt, bitte kontaktiere uns für mehr Information",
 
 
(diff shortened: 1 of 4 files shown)


lib/l10n/app_de.arb
```diff
@@ -19,6 +19,8 @@
   "before": "Vorher",
   "after": "Nachher",
   "failureSnackBar": "Es ist ein Fehler aufgetreten.",
+  "blockedSnackBar": "Dein Account wurde geblockt, bitte kontaktiere uns für mehr Information",
```
fynsta, line 22: Informationen

lib/main.dart
```diff
@@ -83,6 +84,16 @@ class AuthAppState extends State<AuthApp> {
         final AuthChangeEvent event = data.event;
         final Session? session = data.session;
         await supabaseManager.reloadCurrentProfile();
+        if (event == AuthChangeEvent.signedIn) {
+          final bool? isBlocked = await supabaseManager.supabaseClient
+              .rpc('is_blocked', params: <String, dynamic>{'user_id': data.session!.user.id}) as bool?;
+          if (isBlocked != null && isBlocked) {
```
fynsta, line 90:
```suggestion
          if (isBlocked ?? false) {
```

test/widgets/main_test.dart
```diff
@@ -151,6 +153,31 @@ void main() {
       expect(find.byType(MainApp), findsOneWidget);
     });
 
+    testWidgets('It shows WelcomePage after signing in when user is blocked', (WidgetTester tester) async {
```
fynsta, line 156: "It logs user out again after..."

fynsta:
Is there no way to block it straight on login? Login - Logout feels dirty?

istzen:
> Is there no way to block it straight on login? Login - Logout feels dirty?

the only way would be to write a new function for the signup with checks whether a user is blocked before the sign in.

fynsta:
As we said, in the future it would be great if we could somehow make sure that the user does not do anything while he is not logged out yet. But of course I have no idea how to do that, and we don't have any time anymore.

fynsta (approved):

PR #149, opened 2023-03-12
Low hanging fruit

Eddie-42:
Kind of picking up the last easy bits of coverage. Also, I refactored the Profile Page Test because it was inconsistent on GitHub (but, weirdly, not locally). It remains to be seen if that works!

### DIFF
--- a/test/widgets/home_page_test.dart
+++ b/test/widgets/home_page_test.dart
@@ -102,6 +102,95 @@ void main() {
   });
 
   group('RideEvent', () {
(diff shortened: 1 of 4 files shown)


fynsta (approved):

fynsta:
Looks like that _did_ fix the flakiness. I would love a linter rule that forbids random code outside of setUp/tearDown/testWidgets

Showcas (approved):

PR #150, opened 2023-03-13
Test accessibility

Eddie-42:
These are automated tests for accessibility. It checks:
* Whether every tappable object is big enough: That is, at least 48*48 pixels.
* Whether every tappable object has a label or tooltip for Screen Readers.
* Whether the WCAG recommended contrast ratio of at least 4.5/1 is met. For more information, look [here](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html).

I had to change a few things in order to get this to work: 
* First of all, the Write Review Page looks a lot different now. Love it or hate it, but I needed to change it so the tap target guideline was met and at least, now the category texts don't wrap nastily anymore.
* I had to remove the InkWell from the values on the ProfilePage, as they weren't big enough either. I tried making them 48 pixels high, but that looked very bad. Also, I made the tap targets of the change avatar button and the username bigger (the latter results in the edit icon always being on the far right).
* I labeled everything tappable.
* Sometimes, I replaced `...onSurface.withOpacity(0.5)` with `...onSurface.withOpacity(0.6)` in order to meet the contrast guideline.


lib/util/profiles/reviews/custom_rating_bar_size.dart
```diff
@@ -1,21 +1,21 @@
-enum CustomRatingBarSize {
```
fynsta, line 1: How is this whole file changed?

test/util/pump_material.dart
```diff
@@ -2,15 +2,18 @@ import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:motis_mitfahr_app/util/locale_manager.dart';
+import 'package:motis_mitfahr_app/util/theme_manager.dart';
 
-Future<void> pumpMaterial(WidgetTester tester, Widget widget, {NavigatorObserver? navigatorObserver}) async {
+Future<void> pumpMaterial(WidgetTester tester, Widget widget,
+    {NavigatorObserver? navigatorObserver, ThemeData? themeData}) async {
```
fynsta, line 8: trailing comma after themeData

fynsta (approved):

Showcas (approved):

PR #157, opened 2023-03-16
Fix local and utc times in the database

fynsta:


Eddie-42 (approved):

fynsta:
Note: This will introduce some weird behaviour. The server creates the drives in UTC. That means that time change will result in a 1-hour gap between different drives.

image

(time change is on March 26)

